PR #2, opened 2018-03-14
switch to read type

darcatron:
@szabowexler


szabowexler:
Fixed by the PR referenced here.

PR #10, opened 2018-03-26
Support conversations.list

darcatron:
Also switches the `channels.list` api to use `ChannelsListParams` and take the passed filters into account when getting the list


slack-base/src/main/java/com/hubspot/slack/client/models/response/conversations/ConversationListResponseIF.java
```diff
@@ -0,0 +1,24 @@
...
+import com.hubspot.slack.client.models.response.SlackResponse;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ConversationListResponseIF extends SlackResponse {
+  @JsonProperty("channels") // just. wow.
+  List<Conversation> getConversations();
+
+  Optional<ResponseMetadata> getResponseMetadata();
```
szabowexler: This is already part of every response in https://github.com/HubSpot/slack-client/blob/master/slack-base/src/main/java/com/hubspot/slack/client/models/response/SlackResponse.java#L7. No need to explicitly include it here.

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -395,7 +399,7 @@ protected Long getInitialOffset() {
   }
 
   private CompletableFuture<Optional<SlackChannel>> findChannelByName(String name, ChannelsFilter channelsFilter) {
-    return searchNextPage(name, listChannels(channelsFilter).iterator());
+    return searchNextPage(name, listChannels(ChannelsListParams.builder().from(channelsFilter).build()).iterator());
```
szabowexler: nit: when using a builder, split over multiple lines using indentation to improve the legibility of the code. So this should become:

```java
return searchNextPage(
  name, 
  listChannels(
    ChannelsListParams.builder()
      .from(channelsFilter)
      .build()
    ).iterator()
  );
```

PR #11, opened 2018-03-26
support slash commands

darcatron:


slack-base/src/main/java/com/hubspot/slack/client/models/commands/SlashCommandIF.java
```diff
@@ -0,0 +1,29 @@
...
+import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface SlashCommandIF {
```
szabowexler: This looks like it should be an abstract class, so that it can implement the following interfaces:

* `HasChannel`
* `HasUser`

I think it's likely also worth adding a `HasTeam` to capture the "team id" field for other params that apply to teams.

slack-base/src/main/java/com/hubspot/slack/client/models/commands/SlashCommandIF.java
```diff
@@ -0,0 +1,29 @@
...
+  String getToken();
+  String getTeamId();
+  String getTeamDomain();
+  Optional<String> getEnterpriseId();
+  Optional<String> getEnterpriseName();
+  String getChannelId();
+  String getChannelName();
+  String getUserId();
+  @Deprecated
```
szabowexler: Why is this deprecated? Can we include a description of the reason for deprecation/link to a place to read about it?

PR #15, opened 2018-04-02
support adding reactions to messages

darcatron:
Support for [reactions.add](https://api.slack.com/methods/reactions.add)

cc @dylanrb123

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/methods/params/reactions/ReactionsAddParamsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/methods/params/reactions/ReactionsAddParamsIF.java
@@ -2,22 +2,32 @@ package com.hubspot.slack.client.methods.params.reactions;
 
 import java.util.Optional;
 
+import org.immutables.value.Value;
 import org.immutables.value.Value.Immutable;
 
 import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.immutables.validation.ImmutableConditions;
 
 @Immutable
 @HubSpotStyle
 @JsonNaming(SnakeCaseStrategy.class)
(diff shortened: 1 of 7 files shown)


slack-base/src/main/java/com/hubspot/slack/client/methods/params/reactions/ReactionsAddParamsIF.java
```diff
@@ -0,0 +1,23 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ReactionsAddParamsIF {
+ String getName();
+ /**
+  * One of file, file_comment, or the combination of channel and timestamp must be specified.
```
szabowexler: Can we specify this via a [check](https://github.com/HubSpot/hubspot-immutables#validation)? I'm generally a fan of expressing restrictions through code rather than comments.

darcatron: my bad. I meant to add the check. I didn't know we had a convenience method for it though. Included a few tests too

darcatron:
@szabowexler look good to merge in?

slack-base/src/main/java/com/hubspot/slack/client/methods/params/reactions/ReactionsAddParamsIF.java
```diff
@@ -0,0 +1,33 @@
...
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.immutables.validation.ImmutableConditions;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ReactionsAddParamsIF {
+  String getName();
+  /**
+   * One of file, file_comment, or the combination of channel and timestamp must be specified.
```
szabowexler: we can get rid of this now -- it's documented in the check!

PR #16, opened 2018-04-04
switch to iterator

darcatron:
Using iterator is much simpler than trying to debug the completable future recursion

cc @dylanrb123


axiak:
@darcatron I'd definitely avoid making a method return CompletableFuture but be implemented in a blocking fashion.

szabowexler:
This is superseded by https://github.com/HubSpot/slack-client/pull/17, closing in favor of that.

PR #23, opened 2018-04-18
Add open conversation method

zklapow:
This adds support for the [conversation.open](https://api.slack.com/methods/conversations.open) method.

@szabowexler

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/methods/params/conversations/ConversationOpenParamsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/methods/params/conversations/ConversationOpenParamsIF.java
@@ -1,0 +1,37 @@
+package com.hubspot.slack.client.methods.params.conversations;
+
+import java.util.List;
+import java.util.Optional;
+
+import org.immutables.value.Value.Default;
+import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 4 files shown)


slack-base/src/main/java/com/hubspot/slack/client/methods/params/conversations/ConversationOpenParamsIF.java
```diff
@@ -0,0 +1,37 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.google.common.base.Joiner;
+import com.hubspot.immutables.style.HubSpotStyle;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ConversationOpenParamsIF {
```
szabowexler: Could you make this an abstract class, and then implement:

* [`HasChannel`](https://github.com/HubSpot/slack-client/blob/master/slack-base/src/main/java/com/hubspot/slack/client/methods/interceptor/HasChannel.java)
* [`HasCommaSeperatedUserIds`](https://github.com/HubSpot/slack-client/blob/master/slack-base/src/main/java/com/hubspot/slack/client/methods/interceptor/HasCommaSeperatedUserIds.java)

I'm using that to make the intercept stuff sane (it's a bit of a pain when creating the params, but it makes writing interceptors easier).

PR #28, opened 2018-05-03
support WTA in OAuthCredentials

dylanrb123:
Slack's new [Workspace Token Auth](https://api.slack.com/docs/working-with-workspace-tokens) has a slightly different OAuth contract. Mainly, it doesn't send back the `scope` or `user_id` fields so those need to be optional here.

cc @darcatron


darcatron:
lgtm so long as any impact on existing builds were handled

we may want to note that this could cause breakage in the next client release

PR #33, opened 2018-05-20
update max char checks

darcatron:
`500` is not the default max for both `text` and `textarea` fields. This updates the validation to verify `text` < 150 and `textarea` < 3000. I also threw in the checks for min values

Note: [Slack's documention](https://api.slack.com/dialogs) is inconsistent for `text` values. For max_length, they mention 
> Maximum input length allowed for element. Up to 150 characters. Defaults to 150

and then the docs say

> A default value for this field. Up to 500 characters
for `value`

I left it all as 150 just to be safe. Will do a quick local test to see if `500` works for `text` and update this PR


darcatron:
Update: It's 150

PR #36, opened 2018-06-21
support users.conversations

darcatron:
tested and it works fine

cc @szabowexler


slack-base/src/main/java/com/hubspot/slack/client/methods/SlackMethods.java
```diff
@@ -156,6 +156,7 @@
   usergroups_users_list(MethodWriteMode.READ, RateLimitingTiers.TIER_2, JsonStatus.FORM_ENCODING_ONLY),
   usergroups_users_update(MethodWriteMode.WRITE, RateLimitingTiers.TIER_2, JsonStatus.ACCEPTS_JSON),
 
+  users_conversations(MethodWriteMode.WRITE, RateLimitingTiers.TIER_3, JsonStatus.FORM_ENCODING_ONLY),
```
szabowexler: @darcatron really? a new method doesn't support json? :suspect:

darcatron: Yeah, it's unfortunate: https://api.slack.com/methods/users.conversations
api encoding warning

szabowexler: Conversations are still new things to me (get off my laaaaaawwwwnnnnnn) -- but oh well.

Issue #21, opened 2018-04-14
make chat.postMessage "text" optional if attachments exist


darcatron:
Right now the `text` field is always required when using `postMessage`, but it should only be required if the attachment is not present.

https://api.slack.com/methods/chat.postMessage
> Text of the message to send... This field is usually required, unless you're providing only attachments instead.

PR #37, opened 2018-06-21
allow empty text field if attachments are provided

darcatron:
Resolves: https://github.com/HubSpot/slack-client/issues/21

Tested locally and the functionality works. 

I tried having the check within the parent `MessageParams` but I got test failures like:
```
pojosDoDeserializeHasChannelsWithChannelField[class com.hubspot.slack.client.methods.params.chat.ChatUpdateMessageParams](com.hubspot.slack.client.HasChannelDeserializerTest)  Time elapsed: 0.002 sec  <<< ERROR!
java.lang.reflect.InvocationTargetException
```

The current build for this is causing similar problems. It's hitting the `@Check` and erroring on the test

Any thoughts on this @szabowexler ?


szabowexler:
That's pretty strange. The test that's failing there is intended to be a safeguard to ensure that the fields have the correct names (since we can't enforce it at the base class level). Your changes don't seem like they should be related though. I'd see if you can capture exactly where it's failing while running locally, and see what comes out if you serialize that to json.

darcatron:
The issue with the testing is that `Optional` types are ignored and also not handled. `default` methods incur a similar problem. 

I've updated the reflection tests to include a set of `Optional<String>` type methods that should be built and included a conditional clause for them. The existing objects are running fine and the new check works as intended. I've included `*MessageParam` tests to verify this as well

@szabowexler

PR #40, opened 2018-07-23
Added include_locale flag to UserInfoParamsIF and added optional loca…

BWehner:
Added an optional include_locale flag on UserParamsIF and the corresponding optional on SlackUser to allow us to use the optional include_local flag on the slack API (see https://api.slack.com/methods/users.info)

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/users/SlackUserIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/users/SlackUserIF.java
@@ -39,6 +39,9 @@ public interface SlackUserIF extends SlackUserCore {
   @JsonProperty("updated")
   Optional<Integer> getRawUpdated();
 
+  @JsonProperty("locale")
+  Optional<String> getLocale();
+
   @Derived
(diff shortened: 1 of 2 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/users/SlackUserIF.java
```diff
@@ -39,6 +39,9 @@
   @JsonProperty("updated")
   Optional<Integer> getRawUpdated();
 
+  @JsonProperty("locale")
+  Optional<String> getLocale();
```
darcatron: `@JsonProperty` is useful if the method name doesn't match the property name. Since we're using `getLocale` here, it should be serialized into `locale`. So we should be fine without the `@JsonProperty` annotation

BWehner: Awesome, I'll take that out then and test

BWehner:
Closed in favor of equivalent add include local flag PR

PR #45, opened 2018-08-17
Add basic support for uploading files

szabowexler:
Per the description, adds support for calling [`files.upload`](https://api.slack.com/methods/files.upload). This isn't complete (no support for `mime/multipart` right now), but allows a user to upload a block of text (i.e. creating a snippet). The class structure here is permissive, and makes it easy to add additional filetype wrappers (I've added image based (gif) file handles as well as the core text to make the eventual extension to mime simpler). Once I've tested it in the wild I'll merge.

@darcatron


szabowexler:
Tested and confirmed this WAD. Merging for glory.

PR #47, opened 2018-08-24
SlackFileTypes -> SlackFileType

stevegutz:
@szabowexler


szabowexler:
I'm reluctant to do this, just because it will break everything, and it seems a small gain. I'd prefer to just leave it as-is.

stevegutz:
What do you mean "break everything"? Are people accessing this enum directly?

stevegutz:
We haven't cut a release yet and aren't using this internally, so lets go for it.

PR #48, opened 2018-08-24
 Add csv file type

stevegutz:
We can talk about the refactoring I did, but pushing this branch will allow me to start using it.

Depends on https://github.com/HubSpot/slack-client/pull/47

@szabowexler

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
@@ -5,14 +5,14 @@ import java.util.stream.Stream;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 
-public enum SlackFileTypes {
+public enum SlackFileType {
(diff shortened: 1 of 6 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
```diff
@@ -5,14 +5,15 @@
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 
-public enum SlackFileTypes {
+public enum SlackFileType {
   TEXT("text"),
   GIF("gif"),
+  CSV("csv")
```
szabowexler: can we also include a trailing `,`, just so the next addition doesn't touch this line in the dif?

stevegutz: Yes, I  usually do that and I'm sorry I didn't

stevegutz:
This no longer includes https://github.com/HubSpot/slack-client/pull/47, so we can deal with that separately.

PR #50, opened 2018-08-31
Examine headers for potential auth headers, and auto-redact them

szabowexler:
We should carefully never print full auth tokens if we can infer that that's what we're handling.


ssalinas:
🎉

PR #54, opened 2018-09-20
Channel events

BWehner:
Added new Slack events for `channel_created`, `channel_deleted`, `channel_archive`, `channel_unarchive`, and `channel_rename`

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/events/channel/SlackChannelArchiveEventIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/events/channel/SlackChannelArchiveEventIF.java
@@ -19,4 +19,8 @@ public interface SlackChannelArchiveEventIF extends SlackEvent {
 
   @JsonProperty("user")
   String getUserId();
+
+  default String getTs() {
+    return null;
+  }
 }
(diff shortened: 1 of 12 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/events/channel/SlackChannelCreatedEventIF.java
```diff
@@ -0,0 +1,27 @@
...
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonDeserialize(as = SlackChannelCreatedEvent.class)
+public interface SlackChannelCreatedEventIF extends SlackEvent {
+  //Channel.created and channel.creator do not exist on the SlackChannel object and consequently aren't parsed right now. If we need these in the future
+  //we can define a wrapper around SlackChannel containing those two fields.
+
+  SlackChannel getChannel();
+
+  //Channel deleted events do not have a ts, so we manually set it as null
+  default String getTs() {
```
darcatron: > //Channel _deleted_ events

this is the channel _created_ object. Can we update these to all have the correct version of this comment?

slack-base/src/main/java/com/hubspot/slack/client/models/events/channel/SlackChannelArchiveEventIF.java
```diff
@@ -0,0 +1,26 @@
...
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonDeserialize(as = SlackChannelArchiveEvent.class)
+public interface SlackChannelArchiveEventIF extends SlackEvent {
+  @JsonProperty("channel")
+  String getChannelId();
+
+  @JsonProperty("user")
+  String getUserId();
+
+  default String getTs() {
```
darcatron: I think it'd be clearer here if we had the `@Override` annotation on these methods

PR #55, opened 2018-10-01
Add slack event json view

BWehner:
Adding an internal JSON view to SlackEventWrapper to avoid serializing tokens in logs and other places we don't mean to


darcatron:
I might be confused on where this serialization is happening here, but doesn't this prevent anything that sends this wrapper over the wire from having access to the token? Then the receiver can't verify that token is valid

BWehner:
Yep, you're correct. Closing this PR

Issue #52, opened 2018-09-20
ts is required by SlackEvent model, but Slack does not set ts on all events


BWehner:
**Describe the bug**
`ts` is required by the SlackEvent interface, but slack only sets `ts` on some events, such as messages sent, while not on others, such as channel_created

**To Reproduce**
Try to parse a channel_created to a SlackEvent

**Expected behavior**
All slack events JSON should be parseable to SlackEvent

**Additional context**
**Note: Fixing this would be a breaking change, as it would change the SlackEvent interface**
_Edit: Actually, this could be fixed in a non-breaking way by making ts nullable, which would be uglier but would fix this bug_ 
See: https://api.slack.com/events-api#event_type_structure
https://api.slack.com/events/channel_created
https://api.slack.com/events/message (which DOES include ts)

Example channel_created message:
```
{
  "token": "uWopedqK86MJfjUeq8v9pUAD",
  "team_id": "TBQPW3FH6",
  "api_app_id": "A945JNH24",
  "event": {
    "type": "channel_created",
    "channel": {
      "id": "CCXF9BDV1",
      "is_channel": true,
      "name": "test",
      "name_normalized": "test",
      "created": 1537452804,
      "creator": "UBRFHDFGA",
      "is_shared": false,
      "is_org_shared": false
    },
    "event_ts": "1537452804.000100"
  },
  "type": "event_callback",
  "event_id": "EvCXCPK952",
  "event_time": 1537452804,
  "authed_users": [
    "UBRFHDFGA"
  ]
}
```

szabowexler:
This is pretty dangerous. In general, the way to handle this is to change the way it's set up by adding an optional version, migrating usages onto that, and going forward from there. We don't generally want to return `null` to get around this, because while it may not break the API, it violates the invariant that any non-optional return is also non-null.

darcatron:
> adding an optional version

Do you mean to add _another_ field that has an optional timestamp? AFAIK, deserialization can only map a key to a single field. So the new optional field would have to be `Derived` and then later updated once usages are migrated. Is that what you mean?

szabowexler:
Yeah exactly.

PR #56, opened 2018-10-08
User events

BWehner:
Added a user_change events and some tests and deserializing it

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/events/user/SlackUserChangeEventIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/events/user/SlackUserChangeEventIF.java
@@ -1,0 +1,25 @@
+package com.hubspot.slack.client.models.events.user;
+
+import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.events.SlackEvent;
+import com.hubspot.slack.client.models.users.SlackUser;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonDeserialize(as = SlackUserChangeEvent.class)
+public interface SlackUserChangeEventIF extends SlackEvent {
+
+  SlackUser getUser();
+
+  //User change events do not have a ts, so we manually set it as null
+  @Override
+  default String getTs() {
+    return null;
+  }
+}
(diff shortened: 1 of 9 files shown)


BWehner:
cc @darcatron

slack-base/src/test/java/com/hubspot/slack/client/models/events/json/EventDeserializerTest.java
```diff
@@ -80,6 +81,25 @@ public void itCanDeserChannelUnarchivedEvent() throws IOException {
...
+  public void itCanDeserUserChangeEventUsernameChanged() throws IOException {
+    SlackEvent event = fetchAndDeserializeSlackEvent("user_change_username_changed_message.json");
+    assertThat(event.getType()).isEqualTo(SlackEventType.USER_CHANGE);
+  }
+
+  @Test
+  public void itCanDeserUserChangeEventDisplayNameChanged() throws IOException {
+    SlackEvent event = fetchAndDeserializeSlackEvent("user_change_display_name_changed_message.json");
+    assertThat(event.getType()).isEqualTo(SlackEventType.USER_CHANGE);
+  }
```
darcatron: I might be missing something here, but there doesn't seem to be any way to know what, in particular, was changed for the user when the `user_changed` event comes through. So these json files with varying field values don't seem like they're really testing anything.

BWehner: That's exactly what it's testing, yeah. Failing to deserialize the events was what made my initial channels webhooks fail, so this just proves that these events can be properly deserialized

darcatron: So each of these tests have different names and different json files, but as far as the parsing goes, the objects are all the same. The only real (and testable) difference in the files looks to be the `type` field.  Seems like we don't need to test 3 varieties of the user json for the same type?

slack-base/src/main/java/com/hubspot/slack/client/models/events/user/SlackUserChangeEventIF.java
```diff
@@ -0,0 +1,25 @@
...
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonDeserialize(as = SlackUserChangeEvent.class)
+public interface SlackUserChangeEventIF extends SlackEvent {
+
+  SlackUser getUser();
+
+  //User change events do not have a ts, so we manually set it as null
+  @Override
+  default String getTs() {
+    return null;
```
szabowexler: I'm a little concerned about this. What you're really saying is that this is actually nullable, but by doing it this way you've avoided changing the method signature. This exposes the possibility of a NPE down the road, because you've violated the contract. The right answer is to make slack events have optional TS since not every slack event actually has a TS, or to make TS not be a required field on the base `SlackEvent`.

darcatron: Making getTs optional is a breaking change, but removing getTs from the `SlackEvent` so that it's not required sounds like it would achieve the desired effects and circumvent breaking existing usages. Then we can just define getTs in the events that have it. For this repo's purposes, it's just `SlackEventMessage` for `Message` events. I think that makes good sense

BWehner: See discussion on https://github.com/HubSpot/slack-client/issues/52

PR #59, opened 2018-10-24
Fix user deserialization

BWehner:
Fixing an error with user deserialization causing deleted to always be null

cc @darcatron

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/users/SlackUserIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/users/SlackUserIF.java
@@ -15,6 +15,7 @@ import com.hubspot.immutables.style.HubSpotStyle;
 @JsonNaming(SnakeCaseStrategy.class)
 public interface SlackUserIF extends SlackUserCore {
   Optional<UserProfile> getProfile();
+  @JsonProperty("is_deleted")
   Optional<Boolean> isDeleted();
   Optional<String> getColor();
   Optional<String> isAdmin();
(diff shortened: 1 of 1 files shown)


szabowexler:
Could you add a regression test here? Grab an actual slack response, dummy out the PII, and make sure that future changes don't break this?

BWehner:
Regression test added

slack-base/src/test/java/com/hubspot/slack/client/models/users/SlackUserDeserializationTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Test
+  public void serializeAndDeserializeAreInverses() throws IOException {
+    SlackUser user = getSlackUser();
+    SlackUser userDeserialized = ObjectMapperUtils.mapper().readValue(ObjectMapperUtils.mapper().writeValueAsString(user), SlackUser.class);
+    assertThat(userDeserialized).isEqualTo(user);
+  }
+
+  private SlackUser getSlackUser() throws IOException {
+    return ObjectMapperUtils.mapper().readValue(JsonLoader.loadJsonFromFile("deleted_slack_user.json"), new TypeReference<SlackUser>() {});
```
darcatron: Do you need TypeRef here?
```suggestion
    return ObjectMapperUtils.mapper().readValue(JsonLoader.loadJsonFromFile("deleted_slack_user.json"), SlackUser.class);
```

BWehner: Nope, fixed

PR #61, opened 2018-10-29
Added isGeneral flag to SlackChannel 

BWehner:
Also added a test to prove it deserializes properly with isGeneral set


BWehner:
cc @darcatron

Issue #57, opened 2018-10-10
Allow custom rate limiters on SlackClient


BWehner:
Currently, SlackClient only rate limits on method, not the method-token combination that Slack actually rate limits apps on. To fix this, we'd like to be able to supply an optional RateLimiter via SlackRuntimeConfig with some well defined interface, and, if supplied, use that rate limiter in SlackClient rather than the static one (i.e. RATE_LIMITERS). Elias has also suggested we cut a release on this feature

darcatron:
This is out now! Thanks @BWehner!

PR #64, opened 2018-11-01
Allowing custom rate limiters on the slack client

BWehner:
1. Defined a new SlackRateLimiter interface with one method, acquire(String slackToken, SlackMethod slackMethod) which acquires a permit for the given token method combination
2. Defined an implementation ByMethod, which encapsulates the current rate limiting implementation in SlackClient (i.e. the static map of RATE_LIMITERS from method to guave RateLimiter)
3. Adds an option on SlackRuntimeConfig to specify a rate limiter
4. Inject the new ByMethodRateLimiter in SlackWebClient, use it if slackRuntimeConfig.getSlackRateLimiter() is empty

**Testing**
Tested locally by building off branch in slack-integration and running the updateSlackChannels job, which uses Slack Client. Job operated as expected

### DIFF
--- a/slack-java-client/src/main/java/com/hubspot/slack/client/ratelimiting/ByMethodRateLimiter.java
+++ b/slack-java-client/src/main/java/com/hubspot/slack/client/ratelimiting/ByMethodRateLimiter.java
@@ -12,7 +12,7 @@ public class ByMethodRateLimiter implements SlackRateLimiter {
   @Override
   public double acquire(String slackToken, SlackMethod slackMethod) {
     double permissibleQueriesPerSecond = slackMethod.getRateLimitingTier().getMinutelyAllowance() / 60.0;
-    return RATE_LIMITERS.computeIfAbsent( //TODO bwehner this is where we'd call the interface
+    return RATE_LIMITERS.computeIfAbsent(
         slackMethod,
         ignored -> RateLimiter.create(permissibleQueriesPerSecond)
     ).acquire();
--- a/slack-java-client/src/main/java/com/hubspot/slack/client/ratelimiting/ByMethodRateLimiter.java
+++ b/slack-java-client/src/main/java/com/hubspot/slack/client/ratelimiting/ByMethodRateLimiter.java
@@ -1,0 +1,20 @@
+package com.hubspot.slack.client.ratelimiting;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import com.google.common.util.concurrent.RateLimiter;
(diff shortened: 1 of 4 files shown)


BWehner:
See https://github.com/HubSpot/slack-client/issues/57

slack-java-client/src/main/java/com/hubspot/slack/client/ratelimiting/ByMethodRateLimiter.java
```diff
@@ -0,0 +1,20 @@
...
+import com.google.common.util.concurrent.RateLimiter;
+import com.hubspot.slack.client.methods.SlackMethod;
+
+public class ByMethodRateLimiter implements SlackRateLimiter {
+  private static final ConcurrentMap<SlackMethod, RateLimiter> RATE_LIMITERS = new ConcurrentHashMap<>();
+
+  @Override
+  public double acquire(String slackToken, SlackMethod slackMethod) {
+    double permissibleQueriesPerSecond = slackMethod.getRateLimitingTier().getMinutelyAllowance() / 60.0;
+    return RATE_LIMITERS.computeIfAbsent( //TODO bwehner this is where we'd call the interface
```
darcatron: can we remove this todo?

slack-java-client/src/main/java/com/hubspot/slack/client/ratelimiting/ByMethodRateLimiter.java
```diff
@@ -0,0 +1,20 @@
+package com.hubspot.slack.client.ratelimiting;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import com.google.common.util.concurrent.RateLimiter;
+import com.hubspot.slack.client.methods.SlackMethod;
+
+public class ByMethodRateLimiter implements SlackRateLimiter {
```
szabowexler: This should almost certainly be annotated as `@Singleton` to avoid having many instances around (since it's backed by a shared map).

slack-java-client/src/main/java/com/hubspot/slack/client/ratelimiting/ByMethodRateLimiter.java
```diff
@@ -0,0 +1,20 @@
...
+  private static final ConcurrentMap<SlackMethod, RateLimiter> RATE_LIMITERS = new ConcurrentHashMap<>();
+
```
szabowexler: You need to add an `@Inject` constructor here, otherwise guice doesn't know how to put this all together.

PR #65, opened 2018-11-02
Made SlackMemberJoinedChannelEvent deserialize ChannelType to an enum value directly

BWehner:
Also updated SlackChannel to use the new slack channel type name -> ChannelType mapping

**Testing**
Added unit test for deserializing private channel

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/ChannelType.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/ChannelType.java
@@ -1,6 +1,30 @@
 package com.hubspot.slack.client.models;
 
+import java.util.Arrays;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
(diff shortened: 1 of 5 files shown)


BWehner:
cc @darcatron

slack-base/src/main/java/com/hubspot/slack/client/models/ChannelType.java
```diff
@@ -1,6 +1,30 @@
...
+  private final String slackName;
+
+  ChannelType(String slackName) {
+    this.slackName = slackName;
+  }
+
+  @JsonCreator
+  public static ChannelType fromSlackName(String slackName) {
+    return Arrays.stream(ChannelType.values())
+        .filter(enumVal -> enumVal.slackName.toLowerCase().equals(slackName.toLowerCase()))
```
darcatron: should be easier to use `equalsIgnoreCase`

PR #67, opened 2018-11-06
Revert "Allowing custom rate limiters on the slack client"

BWehner:
Reverts HubSpot/slack-client#64


szabowexler:
Thank you -- we'll try the other one again after we add a simple guice verification test to ensure we don't make that same mistake again.

PR #68, opened 2018-11-06
Custom rate limiter

BWehner:
1. Defined a new SlackRateLimiter interface with one method, acquire(String slackToken, SlackMethod slackMethod) which acquires a permit for the given token method combination
2. Defined an implementation ByMethod, which encapsulates the current rate limiting implementation in SlackClient (i.e. the static map of RATE_LIMITERS from method to guave RateLimiter)
3. Adds an option on SlackRuntimeConfig to specify a rate limiter
4. Inject the new ByMethodRateLimiter in SlackWebClient, use it if slackRuntimeConfig.getSlackRateLimiter() is empty

**Testing**
Tested locally by building off branch in slack-integration and running the updateSlackChannels job, which uses Slack Client. Job operated as expected

Also added a test to make sure guice runs correctly, as well as configuring SlackClientModule's binder to require @inject constructors. The reason this merge caused keplar to fail before while slack-integration worked fine with these changes was because some other module in keplar had configured it's own binder to @inject constructors, and that module then installed SlackClientModule. When not configured this way, Guice can still inject instances without a constructor annotated with @inject if the class has a no arg constructor (which guice will use). I Removed @inject annotation form ByMethodRateLimiter and ensured the new Guice test fails, as well as building and running slack-integration off branch to ensure this doesn't break that. Also removed @inject annotation from an injectable class in slack-integration to ensure this change doesn't require anyone who installs this module to have all constructors annotated with @inject, ensuring this is NOT a breaking change 

cc @darcatron @szabowexler


slack-java-client/src/main/java/com/hubspot/slack/client/SlackClientModule.java
```diff
@@ -7,6 +7,8 @@
 public class SlackClientModule extends AbstractModule {
   @Override
   protected void configure() {
+    binder().requireAtInjectOnConstructors();
```
darcatron: this caused issues with any class that has 0 arg constructors. can we drop this line?

szabowexler:
Can you run kepler locally and verify that this behaves itself on this branch?

darcatron:
I verified Kepler and Slack. THose look fine

slack-java-client/src/test/java/com/hubspot/slack/client/SlackClientModuleTest.java
```diff
@@ -0,0 +1,14 @@
+package com.hubspot.slack.client;
+
+import org.junit.Test;
+
+import com.google.inject.Guice;
+import com.google.inject.Stage;
+
+public class SlackClientModuleTest {
+
```
darcatron:
```suggestion
private static class PickyGuiceModule extends AbstractModule {
    @Override
    protected void configure() {
      binder().requireAtInjectOnConstructors();
      install(new SlackClientModule());
    }
  }
```
Can we create the guice module in this test class directly rather than having a new class? It's easier to read.

slack-java-client/src/test/java/com/hubspot/slack/client/SlackClientModuleTest.java
```diff
@@ -0,0 +1,14 @@
...
+  @Test
+  public void itGuices() {
+    Guice.createInjector(Stage.TOOL, new SlackClientTestModule());
```
darcatron:
```suggestion
    Guice.createInjector(Stage.TOOL, new PickyGuiceModule());
```

slack-java-client/src/test/java/com/hubspot/slack/client/SlackClientModuleTest.java
BWehner: Yep, done

PR #69, opened 2018-11-14
Use concrete ChannelsInfoParams

andybergon:
Is there a specific reason why `SlackClient::getChannelInfo` takes an `AbstractChannelsInfoParams` instead of the concrete implementation `ChannelsInfoParams` like all the other methods do? Do we expect people to give it a different implementation?


darcatron:
Nope, you're correct 🙂. That should be `ChannelsInfoParams`. Once you update the import from `import com.hubspot.slack.client.methods.params.channels.AbstractChannelsInfoParams;` to `import com.hubspot.slack.client.methods.params.channels.ChannelsInfoParams;` and grab the latest from master, the build should succeed. We can merge this in after that

darcatron:
Thanks for the PR!

PR #70, opened 2018-11-20
add latest message model to properly deserialize groups

dylanrb123:
The JSON for a latest message is not at all the same as a regular message. I went with a new model instead of changing the existing one because that has usage elsewhere so I didn't want to overload the model and this is quite different. 

@szabowexler

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/EditInfoIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/EditInfoIF.java
@@ -1,0 +1,15 @@
+package com.hubspot.slack.client.models;
+
+import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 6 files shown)


dylanrb123:
@szabowexler mind taking a look at this when you have a chance? Just realized it never got merged

darcatron:
@dylanrb123 have you guys been using these changes at all? I see no problem in merging this in, but it'd be nice to have some extra validation that it's been working for you guys in more realistic use cases

dylanrb123:
No we aren't using it, I was just fixing a bug that someone else reported.

szabowexler:
Sorry, totally missed this. I'm in next week and can take a look then. I'm
also planning on cutting a release, so we can merge and deal with all
outstanding PRs then.

darcatron:
This is stale so I'm going to close this out. If anyone wants to revisit it, the code will be accessible

PR #74, opened 2019-01-09
Add a method for kicking users from a group

axiak:


darcatron:
You can get the best of all worlds (groups, channels, im, and mpim) by using [`conversations.kick`](https://api.slack.com/methods/conversations.kick). 
It'd be awesome if you could add that API method too 😬

PR #79, opened 2019-01-30
Adding state and notify on cancel to dialogs

mlr46:
Adds the state and the notify on cancel field for dialogs.

Documentation is here: https://api.slack.com/dialogs#submit

\cc @szabowexler

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/dialog/SlackDialogIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/dialog/SlackDialogIF.java
@@ -52,18 +52,16 @@ public interface SlackDialogIF {
       throw new IllegalStateException("At most 5 form elements allowed, got " + getElements().size());
     }
 
-    getState()
-        .ifPresent(state -> {
-          if (state.length() > 3000) {
-            throw new IllegalStateException("State cannot exist 3000 chars, got " + state.length());
-          }
-        });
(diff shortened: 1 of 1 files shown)


szabowexler:
Just a quick thing from the docs:

> The state string may be up to 3,000 characters long, and it isn't shown to users

We should add a check to make sure that's not exceeded at runtime, then this lgtm

slack-base/src/main/java/com/hubspot/slack/client/models/dialog/SlackDialogIF.java
```diff
@@ -50,11 +52,16 @@ default void validate() {
       throw new IllegalStateException("At most 5 form elements allowed, got " + getElements().size());
     }
 
-    if (getSubmitLabel().isPresent()) {
-      String label = getSubmitLabel().get();
-      if (label.length() > 24) {
-        throw new IllegalStateException("Submit label cannot exceed 24 chars, got " + label);
+    getState().ifPresent(state -> {
+      if (state.length() > 3000) {
+        throw new IllegalStateException("State cannot exist 3000 chars, got " + state.length());
```
darcatron: > State cannot exist 3000 chars
State cannot **exceed** 3000 chars

mlr46: :+1: will update.

Issue #76, opened 2019-01-28
JSON parsing error in SlackWebClient.shareFilePublically method


alexnivanov:
**Describe the bug**
I'm calling `SlackWebClient.shareFilePublically()` method which executes successfully (I get correct JSON in the log), however it fails with JSON parsing error:

```
2019-01-28 13:26:55 +0100 [ERROR] from com.hubspot.slack.client.interceptors.http.DefaultHttpResponseDebugger in NioHttpClient-Callback-1 - REQ<1> [files_sharedPublicURL]: Failed interaction
POST https://slack.com/api/files.sharedPublicURL
------------------------------------------
Authorization = Bearer xox...269dContent-Type = application/json
------------------------------------------
{
  "file" : "FFS6VUKPG"
}
------------------------------------------
Content-Type = application/json; charset=utf-8
Connection = keep-alive
Date = Mon, 28 Jan 2019 12:26:55 GMT
Server = Apache
Expires = Mon, 26 Jul 1997 05:00:00 GMT
X-Content-Type-Options = nosniff
Pragma = no-cache
Vary = Accept-Encoding
X-Accepted-OAuth-Scopes = files:write:user
X-OAuth-Scopes = identify,bot,commands,channels:history,files:write:user
Cache-Control = private, no-cache, no-store, must-revalidate
X-XSS-Protection = 0
X-Slack-Req-Id = 1022cec4-d6a2-4b67-bb52-b727c610b5e6
Strict-Transport-Security = max-age=31536000; includeSubDomains; preload
Referrer-Policy = no-referrer
Access-Control-Allow-Origin = *
X-Via = haproxy-www-xkzu
X-Cache = Miss from cloudfront
Via = 1.1 59574f77a7cf2d23d64904db278e5711.cloudfront.net (CloudFront)
X-Amz-Cf-Id = Rc75U-ml_lCJCy9rPy7fP2MbP51xlAtm_OYxeFoaZZYM_qqoamtWuQ==
Content-Length = 2380
------------------------------------------
{
  "ok" : true,
  "file" : {
    "id" : "FFS6VUKPG",
    "created" : 1548678406,
    "timestamp" : 1548678406,
    "name" : "1check8-02.jpg",
    "title" : "1check8-02.jpg",
    "mimetype" : "image/jpeg",
    "filetype" : "jpg",
    "pretty_type" : "JPEG",
    "user" : "UCXD8T8SG",
    "editable" : false,
    "size" : 372643,
[...]

PR #81, opened 2019-01-31
Adding JPG file type

mlr46:
\closes https://github.com/HubSpot/slack-client/issues/76

However, we would want to maybe consider a different implementation in order to add all the file types here https://api.slack.com/types/file ?

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
@@ -9,6 +9,7 @@ public enum SlackFileType {
   TEXT("text"),
   GIF("gif"),
   CSV("csv"),
+  JPG("jpg")
   ;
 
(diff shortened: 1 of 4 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
```diff
@@ -9,6 +9,7 @@
   TEXT("text"),
   GIF("gif"),
   CSV("csv"),
+  JPG("jpg")
```
darcatron:
```suggestion
  JPG("jpg"),
```
nit: comma at the end helps keep future diffs clean

szabowexler:
This seems wrong. Jpg is an image file type, so probably shouldn't be inheriting the text file type core, no?

mlr46:
Yah, it should probably inherit from whatever the gif file is inheriting from. :copy-pasta:

PR #85, opened 2019-04-26
handle "errors" instead of "error" in response

wsorenson:
Sometimes the slack API is returning a list of `errors` instead of `error`. While not perfect, this causes the client to not break on that response.


wsorenson:
Response from Slack on the topic:

> Thank you for your patience. I spoke with one of the API engineers, and this is a discrepancy that is expected, but also something that they intend to fix in a future release. The desired behavior is to have error and errors for both public and private channels.
>
> While I'm not able to give you an exact date for when this will be released, it is expected to be relatively soon.
>
>If you're able to handle both in the code, I recommend doing so for the short term in case the true resolution is delayed for any reason.
>
> Thanks again. I appreciate you bringing this discrepancy to our attention.

PR #87, opened 2019-05-07
add conversations reply api

wsorenson:


slack-base/src/main/java/com/hubspot/slack/client/methods/params/conversations/ConversationsRepliesParamsIF.java
```diff
@@ -0,0 +1,17 @@
...
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ConversationsRepliesParamsIF {
+  Optional<String> getChannel();
```
szabowexler: Reading https://api.slack.com/methods/conversations.replies, it looks like `channel` is a required field.

slack-base/src/main/java/com/hubspot/slack/client/methods/params/conversations/ConversationsRepliesParamsIF.java
```diff
@@ -0,0 +1,17 @@
...
+  Optional<String> getTs();
```
szabowexler: Reading https://api.slack.com/methods/conversations.replies, it looks like `ts` is a required field

Issue #83, opened 2019-02-13
This project is way too complicated. Can you include a plain Java hello world without all the annotations and injection?


jackmahoney:
Factories, factories, factories. This project is so complicated. I gave up trying after looking at the source code and trying to get this to work in a plain Java environment without Guice or DI. Please consider including a simple HelloWorld that posts to a slack channel.

latos:
There's nothing wrong with Guice itself, but a well designed library should't impose it (or its own http client) on other developers.

It would be better to have a pure library that just contains the code, with examples for wiring it up manually.  Separately, a library containing a Guice module can be provided for the convenience of those who are using Guice in their projects.  (DI libraries in general make sense as top level choices for applications, they shouldn't be imposed from the bottom up from "leaf" libraries).

namedgraph:
My thoughts exactly. Take a look at google-api-java-client for a simple design.

abhidemon:
This is the reason why this project could have 100 stars by now, is stuck at 40.

kmclarnon:
This is now available in [Release 1.2](https://github.com/HubSpot/slack-client/releases/tag/1.2)

kennycason:
I know this is old, but I'd also like to see an example in the README for how to configure this without Guice. Of course, I'm going to just dig through all the stuff that is configured by Guice to see how it's supposed to be configured. But after 15 min of looking I still don't have a clear way on a clean/easy way to instantiate things. 

I'd expect something simple like:
```kotlin
val slackClient = SlackWebClient(config);
```

PR #88, opened 2019-05-28
Make guice usage optional

kmclarnon:
Our open source projects should not depend on guice.  This PR removes the guice dependency and related factory / abstraction layers and re-implements them in standard java.  As we use guice extensively internally at HubSpot, this PR also provides a bridge module that will ensure the appropriate bindings are in place.

Standard usage of the client looks like this

```java
// define your runtime Config
SlackClientRuntimeConfig config = SlackClientRuntimeConfig.builder()
        .setTokenSupplier(() -> "a fake token")
        .build()

// use the default client to get a slack client instance
SlackClient client = SlackClientFactory.defaultFactory().build(config);

```

Resolves https://github.com/HubSpot/slack-client/issues/83

@szabowexler @jhaber @stevegutz


darcatron:
Thanks for doing this, Kevin! Mind adding that standard java example to the root readme under "Usage"?

stevegutz:
Moving the module is a breaking change, so we need to be careful:

1. Deprecate`com.hubspot.slack.client.SlackClientModule` and have `com.hubspot.slack.client.guice.SlackClientModule` delegate to that
2. Release
3. Update internal usage to `com.hubspot.slack.client.guice.SlackClientModule`
4. Delete `com.hubspot.slack.client.SlackClientModule`
5. Release

vladmm:
What is the status of this PR? Can it be merged? Looking forward to using the library with this improvement.

kmclarnon:
I admittedly forgot about this PR with all the internal things I need to do, but I'll see if I can get it to a workable place today

kmclarnon:
@stevegutz I resolved the binary compatibility issues.  I'll plan to merge this monday and do a release unless you thing we should change anything

slack-java-client-guice/src/main/java/com/hubspot/slack/client/guice/SlackClientModule.java
```diff
@@ -0,0 +1,29 @@
...
+import com.hubspot.slack.client.http.NioHttpClientFactory;
+
+public class SlackClientModule extends AbstractModule {
+
+  @Override
+  protected void configure() {
+    bind(SlackWebClient.Factory.class).toInstance(SlackClientFactory.defaultFactory());
+    bind(SlackClientFactory.class).toInstance(SlackClientFactory.defaultFactory());
+
+    bind(NioHttpClientFactory.class).toInstance(NioHttpClientFactory.defaultFactory());
```
stevegutz: I think that if this doesn't install the old `SlackClientModule` we might end up with duplicate binding errors when both are installed. Would it be possible to add a test for this case and update this code if there is in fact an error?

slack-java-client/src/main/java/com/hubspot/slack/client/SlackClientFactory.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * Use {@link SlackClientFactory#create(SlackClientRuntimeConfig)} instead
+     */
+    @Override
+    @Deprecated
+    public SlackWebClient build(SlackClientRuntimeConfig config) {
+      return new SlackWebClient(NioHttpClientFactory.defaultFactory(), config);
+    }
```
jhaber: Is there a reason for this method to exist? Since this is new code can we just make everyone call `create` instead?

kmclarnon: The only reason it exists is to have one implementation of the `SlackClientFactory`, which extends the existing `@AssistedIInject` interface.  Keeping this allows us to bind only these implementations without worrying that we will break code that hasn't migrated yet

jhaber: Ah ok I missed the `@Override` at first. In the future will that stuff move to the Guice module? Or just get replaced by something else?

kmclarnon: The plan is to release this code, migrate our internal version to this and start moving people to the new methods and the new `slack-java-client-guice` module.  Then we can delete the `@AssistedInject` factory interface in the `SlackWebClient`, remove this method and drop the slack client module from `slack-java-client` along with all guice-related dependencies

jhaber: Sounds good 👍

stevegutz:
🚢 once built. Thanks!

PR #91, opened 2019-06-13
Fix the build

jhaber:
@szabowexler @darcatron

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.hubspot</groupId>
     <artifactId>basepom</artifactId>
-    <version>18.3</version>
+    <version>25.4</version>
   </parent>
 
   <groupId>com.hubspot.slack</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -5,27 +5,42 @@
...
-  </properties>
-
   <modules>
     <module>slack-base</module>
     <module>slack-java-client</module>
     <module>slack-java-client-examples</module>
   </modules>
 
+  <properties>
+    <dep.immutables.version>2.2.10</dep.immutables.version>
```
jhaber: Need to downgrade Immutables because newer versions don't like `SlackEvent#toDetailedEvent`:
```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /Users/jhaber/src/slack-client/slack-base/src/main/java/com/hubspot/slack/client/models/events/SlackEvent.java:[11,36] Method 'toDetailedEvent' cannot have own generic type parameters. Attribute accessors can only use enclosing type's type variables
```

I've deprecated this method and we should remove it in the future (I don't see any annotation that would cause immutables to ignore it)

PR #95, opened 2019-06-28
These are not always present on files

szabowexler:
Internal testing has found some cases where these fields are not present on the response payload.


szabowexler:
Verified in our production environment.

PR #101, opened 2019-07-22
Feature request: conversation members

Ulya0302:


Ulya0302:
Add support for getting channel members like "conversations.members"

slack-base/src/main/java/com/hubspot/slack/client/models/response/conversations/ConversationMemberResponseIF.java
```diff
@@ -0,0 +1,15 @@
...
+import com.hubspot.slack.client.models.response.SlackResponse;
+import org.immutables.value.Value.Immutable;
+
+import java.util.List;
+
+@Immutable
+@HubSpotStyle
+public interface ConversationMemberResponseIF extends SlackResponse {
+    @JsonProperty("members")
+    List<String> getMember();
```
darcatron: Since this API returns a list of IDs, it'll be more clear to have the method named `getMemberIds`

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -716,6 +718,45 @@ protected Long getInitialOffset() {
...
+    return new AbstractPagedIterable<Result<List<String>, SlackError>, String>() {
+      @Override
+      protected String getInitialOffset() {
+        return null;
+      }
+
+      @Override
+      protected LazyLoadingPage<Result<List<String>, SlackError>, String> getPage(String offset) throws Exception {
+        if (LOG.isTraceEnabled()) {
+          LOG.trace("Fetching slack usergroup page from {}", offset);
```
darcatron: > Fetching slack conversations member page...

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -716,6 +718,45 @@ protected Long getInitialOffset() {
...
+        }
+
+        CompletableFuture<Result<ConversationMemberResponse, SlackError>> resultFuture = postSlackCommand(
+                SlackMethods.conversations_members,
+                ConversationMemberParams.builder()
+                        .from(params)
```
darcatron: We'll want to set the `cursor` to the `offset` if it exists. Otherwise, we'll always grab the first page

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -716,6 +718,45 @@ protected Long getInitialOffset() {
...
+                        .build(),
```
darcatron: We have a pattern for paged APIs where we configure the default batch size (`limit`) in the `SlackClientRuntimeConfig`. 

https://github.com/HubSpot/slack-client/blob/25232a17884708c860880b917e8dafaf89fb6ea7/slack-java-client/src/main/java/com/hubspot/slack/client/SlackClientRuntimeConfigIF.java#L27-L30


Right now, none of our Paged methods utilize the caller's `limit`. I think it makes sense to take the user passed in `limit` if it exists. If it doesn't exist, we should fall back to the default `SlackClientRuntimeConfig` value

e.g.
```java
ConversationMemberParams.builder builder = ConversationMemberParams.builder()
                        .from(params);
if (!params. getLimit().isPresent()) {
    builder.setLimit(config.getConversationsMembersBatchSize().get())
}
...
```

mindspin311: Agreed. Plus Slack has really been pushing for 1000 limit pages in many of their API calls, so I can see them rolling that out for everything.

Ulya0302:
I'm sorry for delaying in correcting too

slack-base/src/main/java/com/hubspot/slack/client/methods/params/conversations/ConversationMemberParamsIF.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Optional;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public interface ConversationMemberParamsIF {
+    Optional<String> getCursor();
+    Optional<Integer> getLimit();
+    @JsonProperty("channel")
+    Optional<String> getChannelId();
```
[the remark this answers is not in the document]
Ulya0302: Yes, a looked in the slack docs and we can remove ```Optional```

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
Ulya0302: Ooops)

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
Ulya0302: I hope I undestood correctly this

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
Ulya0302: When I write this I focused on method "listUsergroups" that have same realisation (same mistakes: no offset and no batch size)
Now, I try to look difference in slack docs and don't find information about paging in [usergroup_list
](https://api.slack.com/methods/usergroups.list)

darcatron: Yeah, there are some inconsistencies in the APIs unfortunately

mindspin311:
👍

PR #102, opened 2019-08-14
allow proactive rate limiting

darcatron:
The current rate limiter implementation asks, "when can we send a request" rather than "should we send a request." There's no easy way to proactively stop a request from going out if we know it'll fail. With the current implementation, we must either wait until the request can be sent (bad for high throughput services) or throw an exception (not ideal).

This PR adds support for returning a value to indicate the request was proactively rate limited. Proactively rate limiting means the request won't go out to Slack and the method caller can handle the rate limit just as they would have if Slack had actually rate limited.

@mindspin311 
@igorushakov
cc @szabowexler

### DIFF
--- a/slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
+++ b/slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
@@ -142,6 +142,7 @@ import com.hubspot.slack.client.ratelimiting.SlackRateLimiter;
 
 public class SlackWebClient implements SlackClient {
   public static final int RATE_LIMIT_SENTINEL_VALUE = -1;
+  public static final int RATE_LIMIT_LOG_WARNING_THRESHOLD_SECONDS = 5;
 
   private static final Logger LOG = LoggerFactory.getLogger(SlackWebClient.class);
   private static final HttpConfig DEFAULT_CONFIG = HttpConfig.newBuilder()
@@ -1018,7 +1019,7 @@ public class SlackWebClient implements SlackClient {
 
   private double acquirePermit(SlackMethod method) {
     double acquireSeconds = getSlackRateLimiter().acquire(config.getTokenSupplier().get(), method);
-    if (acquireSeconds > 5.0) {
+    if (acquireSeconds > RATE_LIMIT_LOG_WARNING_THRESHOLD_SECONDS) {
       LOG.warn("Throttling {}, waited {} seconds to acquire permit to run", method, acquireSeconds);
     }
 
--- a/slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
+++ b/slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
@@ -17,6 +17,7 @@ import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
 import com.google.common.base.Strings;
 import com.google.common.collect.Multimap;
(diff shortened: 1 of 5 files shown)


slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -930,24 +933,27 @@ protected String getInitialOffset() {
...
     long requestId = REQUEST_COUNTER.getAndIncrement();
-    return executeLogged(requestId, method, request)
+    requestDebugger.debug(requestId, method, request);
+
+    Stopwatch timer = Stopwatch.createStarted();
+    double acquireSeconds = acquirePermit(method);
+
+    if (acquireSeconds == RATE_LIMIT_SENTINEL_VALUE) {
+      responseDebugger.debugProactiveRateLimit(requestId, method, request);
+      return CompletableFuture.completedFuture(Result.err(SlackError.of(SlackErrorType.RATE_LIMITED.key())));
```
darcatron: I utilized the existing `SlackErrorType` rather than creating a new error because the rate limit result is intended to be the same whether it came from Slack or was calculated by a configured rate limiter.

slack-java-client/src/test/java/com/hubspot/slack/client/SlackWebClientTest.java
```diff
@@ -0,0 +1,35 @@
...
+import com.hubspot.slack.client.models.response.SlackResponse;
+
+public class SlackWebClientTest {
+
+  @Test
+  public void itProactivelyRateLimits() {
+    SlackWebClient slackClient = (SlackWebClient) SlackClientFactory.defaultFactory().create(SlackClientRuntimeConfig.builder()
+            .setTokenSupplier(() -> "xoxp-test-token")
+            .setRequestDebugger((requestId, method, request) -> {})
+            .setSlackRateLimiter((slackToken, slackMethod) -> SlackWebClient.RATE_LIMIT_SENTINEL_VALUE)
```
darcatron: This isn't meant to be a thorough test for sending requests. It's just a sanity check to ensure we have something checking for the rate limit sentinel value

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -998,11 +1016,13 @@ protected String getInitialOffset() {
     return responseFuture;
   }
 
-  private void acquirePermit(SlackMethod method) {
-    double acquireTime = getSlackRateLimiter().acquire(config.getTokenSupplier().get(), method);
-    if (acquireTime > 5.0) {
-      LOG.warn("Throttling {}, waited {} seconds to acquire permit to run", method, acquireTime);
+  private double acquirePermit(SlackMethod method) {
+    double acquireSeconds = getSlackRateLimiter().acquire(config.getTokenSupplier().get(), method);
```
mindspin311: +1

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -998,11 +1016,13 @@ protected String getInitialOffset() {
...
+    if (acquireSeconds > 5.0) {
```
mindspin311: Can we make this configurable? Or at the very least a constant?

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -930,24 +933,27 @@ protected String getInitialOffset() {
...
       Class<T> responseType
   ) {
     long requestId = REQUEST_COUNTER.getAndIncrement();
-    return executeLogged(requestId, method, request)
+    requestDebugger.debug(requestId, method, request);
+
+    Stopwatch timer = Stopwatch.createStarted();
+    double acquireSeconds = acquirePermit(method);
+
+    if (acquireSeconds == RATE_LIMIT_SENTINEL_VALUE) {
```
mindspin311: Looks good.

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
darcatron: I made it a constant. Making it configurable is out of scope and should be in a different PR if people need it

Issue #58, opened 2018-10-11
Add test to verify method rate limit "tiers" are up to date


darcatron:
### Problem
If the declared rate limit tier for a method is changed by Slack, we don't know our tier is out of date. If they increase the tier, we'll be throttling ourselves unnecessarily. If they reduce the tier, we'll likely to hit rate limits (essentially the in house rate limiter is useless if they reduce the tier)

### Solution
We can write a test to loop through each method and check against Slack's site to verify the tier is correct. The list the tier in the `Facts` section for each method. eg. https://api.slack.com/methods/channels.create. 

This isn't ideal, but I couldn't think of another way to verify this information except checking manually.

PR #103, opened 2019-08-20
correct methods and tiers

darcatron:
I wrote up a job to resolve https://github.com/HubSpot/slack-client/issues/58. The first run lead to these fixes. We'll continue running it internally and make updates as necessary.

Note: we only check what's already defined in the enum list. We don't check new methods Slack may have added that aren't declared on the enum list. 

@mindspin311 @igorushakov

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/methods/SlackMethods.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/methods/SlackMethods.java
@@ -6,8 +6,8 @@ package com.hubspot.slack.client.methods;
 public enum SlackMethods implements SlackMethod {
   api_test(MethodWriteMode.READ, RateLimitingTiers.TIER_4, JsonStatus.FORM_ENCODING_ONLY),
 
-  apps_permission_info(MethodWriteMode.READ, RateLimitingTiers.TIER_2, JsonStatus.FORM_ENCODING_ONLY),
-  apps_permission_request(MethodWriteMode.WRITE, RateLimitingTiers.TIER_2, JsonStatus.FORM_ENCODING_ONLY),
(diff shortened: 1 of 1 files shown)


slack-base/src/main/java/com/hubspot/slack/client/methods/SlackMethods.java
```diff
@@ -59,17 +59,15 @@
...
   dnd_endSnooze(MethodWriteMode.WRITE, RateLimitingTiers.TIER_2, JsonStatus.ACCEPTS_JSON),
   dnd_info(MethodWriteMode.READ, RateLimitingTiers.TIER_3, JsonStatus.FORM_ENCODING_ONLY),
   dnd_setSnooze(MethodWriteMode.WRITE, RateLimitingTiers.TIER_2, JsonStatus.FORM_ENCODING_ONLY),
   dnd_teamInfo(MethodWriteMode.READ, RateLimitingTiers.TIER_2, JsonStatus.FORM_ENCODING_ONLY),
 
   emoji_list(MethodWriteMode.READ, RateLimitingTiers.TIER_2, JsonStatus.FORM_ENCODING_ONLY),
 
-  files_comments_add(MethodWriteMode.WRITE, RateLimitingTiers.TIER_2, JsonStatus.ACCEPTS_JSON),
   files_comments_delete(MethodWriteMode.WRITE, RateLimitingTiers.TIER_2, JsonStatus.ACCEPTS_JSON),
-  files_comments_edit(MethodWriteMode.WRITE, RateLimitingTiers.TIER_2, JsonStatus.ACCEPTS_JSON),
```
darcatron: Slack no longer has these methods

PR #104, opened 2019-08-30
Handling removal of the  property on messages

mindspin311:
Slack id deprecating `replies` on messages. See [April 2019 changelog](https://api.slack.com/changelog)

I audited our repos that use `slack-client` and didn't see anyone actually call `getReplies()`, so I think that just having it return an empty array is the way to go. The new methods are now included and can be used going forward.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/LiteMessageIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/LiteMessageIF.java
@@ -1,10 +1,13 @@
 package com.hubspot.slack.client.models;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
+import org.immutables.value.Value.Derived;
 import org.immutables.value.Value.Immutable;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
@@ -16,21 +19,45 @@ import com.hubspot.slack.client.models.files.SlackFile;
 @JsonNaming(SnakeCaseStrategy.class)
 public interface LiteMessageIF {
   String getType();
+
   Optional<String> getSubtype();
 
   Optional<String> getUser();
+
   Optional<String> getBotId();
+
   Optional<String> getUsername();
+
   String getText();
 
   List<Attachment> getAttachments();
+
   List<SlackFile> getFiles();
 
   @JsonProperty("ts")
   String getTimestamp();
+
   @JsonProperty("thread_ts")
   Optional<String> getThreadTimestamp();
 
   Optional<Integer> getReplyCount();
-  List<ReplySkeleton> getReplies();
+
+  Optional<List<String>> getReplyUsers();
+
+  Optional<Integer> getResplyUsersCount();
+
+  @JsonProperty("latest_reply")
+  Optional<String> getLatestReplyTimestamp();
+
+  /**
+   * @deprecated use {@link #getReplyUsers()} or {@link #getLatestReplyTimestamp()}
+   * These can be used to find the user ids and the last reply timestamp.
+   * Used to return a list of `ReplySkeleton`
+   */
+  @Deprecated
+  @Derived
+  @JsonIgnore
+  default List getReplies() {
+    return Collections.emptyList();
+  }
 }
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/events/util/AbstractSlackReplyMessage.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/events/util/AbstractSlackReplyMessage.java
@@ -1,10 +1,14 @@
 package com.hubspot.slack.client.models.events.util;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 import org.immutables.value.Value.Default;
+import org.immutables.value.Value.Derived;
 import org.immutables.value.Value.Immutable;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
@@ -26,10 +30,27 @@ public abstract class AbstractSlackReplyMessage extends SlackEventMessageBase im
   @JsonProperty("channel")
   public abstract String getChannelId();
 
-  public abstract List<Reply> getReplies();
-
   @Default
   public int getReplyCount() {
     return 0;
+  }
+
+  public abstract Optional<List<String>> getReplyUsers();
+
+  public abstract Optional<Integer> getResplyUsersCount();
+
+  @JsonProperty("latest_reply")
+  public abstract Optional<String> getLatestReplyTimestamp();
+
+  /**
+   * @deprecated use {@link #getReplyUsers()} or {@link #getLatestReplyTimestamp()}
+   * These can be used to find the user ids and the last reply timestamp.
+   * Used to return a list of `ReplySkeleton`
+   */
+  @Deprecated
+  @Derived
+  @JsonIgnore
+  public List getReplies() {
+    return Collections.emptyList();
   }
 }
--- a/README.md
+++ b/README.md
@@ -22,6 +22,18 @@ An asychronous HTTP client wrapping Slack's [RPC-style web api](https://api.slac
 * Provide per-method in-memory rate limiting so you don't have to worry about overwhelming slack from a single process
 * Expose highly configurable hooks to allow custom rate limiting, method filtering, and debugging in an extensible way
 
+
+## Breaking Changes
+
+**NOTICE:**
+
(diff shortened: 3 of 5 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/LiteMessageIF.java
```diff
@@ -15,22 +20,50 @@
...
+  @JsonProperty("latest_reply")
+  Optional<String> getLatestReplyTimestamp();
+
+  /**
+   * @deprecated use {@link #getReplyUsers()} or {@link #getLatestReplyTimestamp()}
+   * These can be used to find the user ids and the last reply timestamp.
+   * Used to return a list of `ReplySkeleton`
+   */
+  @Deprecated
+  @Derived
```
darcatron: Do we want this to be `Derived`? If Slack is sending the value, we should return it and, if not, we can default to the empty list. This gives users a backwards compatible client

slack-base/src/main/java/com/hubspot/slack/client/models/events/util/AbstractSlackReplyMessage.java
```diff
@@ -26,10 +30,27 @@
...
+  @JsonProperty("latest_reply")
+  public abstract Optional<String> getLatestReplyTimestamp();
+
+  /**
+   * @deprecated use {@link #getReplyUsers()} or {@link #getLatestReplyTimestamp()}
+   * These can be used to find the user ids and the last reply timestamp.
+   * Used to return a list of `ReplySkeleton`
+   */
+  @Deprecated
+  @Derived
```
darcatron: Same comment as above about `Derived` vs `Default`

slack-base/src/main/java/com/hubspot/slack/client/models/events/util/AbstractSlackReplyMessage.java
```diff
@@ -26,10 +30,27 @@
   @JsonProperty("channel")
   public abstract String getChannelId();
 
-  public abstract List<Reply> getReplies();
```
darcatron: Slack docs don't mention if the update applies to the [message_replied](https://api.slack.com/events/message/message_replied) event. I reached out to them and they said

> ... our team will be doing some testing to confirm this for you and will send you an update...

If they don't update the `message_replied` event, we won't need to (and shouldn't) update this model

slack-base/src/main/java/com/hubspot/slack/client/models/LiteMessageIF.java
```diff
@@ -15,22 +20,50 @@
...
+
   @JsonProperty("thread_ts")
   Optional<String> getThreadTimestamp();
 
   Optional<Integer> getReplyCount();
-  List<ReplySkeleton> getReplies();
+
+  Optional<List<String>> getReplyUsers();
+
+  Optional<Integer> getResplyUsersCount();
```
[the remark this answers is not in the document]
mindspin311: 🤦‍♂

slack-base/src/main/java/com/hubspot/slack/client/models/LiteMessageIF.java
mindspin311: If you're upgrading to this client, I like that it's always empty.
Rather than have something continue to work and just stop at some point in the future.

I'm also gonna remove `ReplySkeleton` as it probably has no reason to still be here.

slack-base/src/main/java/com/hubspot/slack/client/models/events/util/AbstractSlackReplyMessage.java
mindspin311: How did you reach out to them? I don't see an email or anything in our hubspot-slack Workspace.

darcatron: I just [opened a support ticket](slack.com/help/requests/new). I got a response from them:
```
Hi there,

Thanks for getting in touch. The message_replied event subtype will not be affected by the changelog entry that you mentioned. The message_replied subtype simply indicates that a new message has been posted, and it's a "reply" to another message.

That event will not contain other threaded replies, so you would need to use the conversations.replies API method to fetch those: https://api.slack.com/methods/conversations.replies.
```

So it looks like this file doesn't need to be modified

mindspin311: sounds good. I'll revert that.

slack-base/src/main/java/com/hubspot/slack/client/models/LiteMessageIF.java
mindspin311: Actually looks like `LiteMessage` is used in other places, so probably safe to just do what you suggested. If they are sending it in other places, it will still be used.

darcatron: The goal of this update is to migrate off of `getReplies`. As written, if a user updates to this version, the `getReplies` method still exists in the model. If users try to use it, they either get what they except (before Oct 18th) or get back incorrect data (empty list). 

The fallback to the empty list is not accurate and can lead to unintended logic. This type of deprecation isn't a great experience. If we leave in this deprecated functionality, we should take the side of failing fast to prevent potentially costly misuses. 

I see 2 options

1) We remove the `getReplies` method entirely so users never accidentally call it. Anyone upgrading will have to migrate off `getReplies` since old code won't compile. This is explicit and removes any chance of errors. 

2) We expose a deprecated `getReplies` and log an ERROR whenever this method is called. If Slack no longer supplies this field and `getReplies` is called, we throw a RuntimeException. This will be harder to do with an immutable, but can be useful as users transition off `getReplies`.

I prefer option 1.

mindspin311: So the issue is that `LiteMessage` is part of tons of other model objects. It looks like it's not just for conversations. Releasing this and having it log an error along with marking it as `@Deprecated` seemed like the safest thing to do here.

I'm gonna reach out to Slack and see if it's just the one specific place when https://api.slack.com/methods/conversations.replies is called where replies is being removed, or if it's other places. It may be that our `LiteMessage` needs to be forked for the `conversations.replies` case and all others.

mindspin311: https://hubspot.slack.com/help/requests/2419401 They think this is ONLY `conversations.replies`, but waiting to have them confirm in case there are other places (beyond the `message_replied` event)

slack-base/src/main/java/com/hubspot/slack/client/models/ConversationsRepliesMessageIF.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.files.SlackFile;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ConversationsRepliesMessageIF {
```
mindspin311: Reason for this is that Slack has said the `replies` deprecation is only applicable to the `conversations.replies` endpoint, so we're narrowly making the change here. `LiteMessage` is used in several places, so I didn't want to make a change that could adversely affect the other places.

mindspin311:
I'm working on testing the `conversations.replies` change in a new debug endpoint, then I can do some unit tests with a few different payloads to make sure it all works.

Assuming Slack gets back to me and says that's the only place I need to change, we should be good.

mindspin311:
Tested locally and it worked fine.

Going to check in 2 internal debug endpoints that should be useful in debugging messages.

https://tools.hubteamqa.com/api/get/local.hubspotqa.com%2Fslack-integration%2Fv1%2Fdebug%2Fconversations-history%3FportalId%3D99981638%26channel%3DCMU5G0JMP%26ts%3D1567610406.000400
Was not affected and still returns the `replies` list.

https://tools.hubteamqa.com/api/get/local.hubspotqa.com%2Fslack-integration%2Fv1%2Fdebug%2Fconversations-replies%3FportalId%3D99981638%26channel%3DCMU5G0JMP%26ts%3D1567610406.000400
Was changed and returns new fields, but not the `replies` property.

It also appears nothing actually calls our `getConversationReplies` client method, so we shouldn't have a problem.

slack-base/src/main/java/com/hubspot/slack/client/models/LiteMessageIF.java
mindspin311: Ok, now they say it's everything that returns this `replies` object, so back to `LiteMessage` where the change is.

mindspin311: And made the default an empty array just to be safe.. So essentially back to what we had originally done.

slack-base/src/main/java/com/hubspot/slack/client/models/events/util/AbstractSlackReplyMessage.java
mindspin311: Slack came back and said all events are affected as well, so asked them again to clarify based on this specific example. Given what they come back with, I'll update this to have a default of empty array too
https://hubspot.slack.com/help/requests/2419401

Issue #89, opened 2019-06-06
Support for new Layout (Block Kit Builder friendly)


UncleFirefox:
**Is your feature request related to a problem? Please describe.**
Slack released a new way of building messages. As far as I know this library does not support it yet, or am I wrong?

**Describe the solution you'd like**
Have a builder patter the same way we have it now with ChatPostMessageParams.builder which contemplates the new format (maybe a new type?)

**Describe alternatives you've considered**
Doing all the communication using a simple RestTemplate but I think this library is too good not to have the new layout. https://github.com/seratch/jslack might the option to to go

**Additional context**
Library now supports this: https://api.slack.com/docs/messages/builder
And should be able to support this: https://api.slack.com/tools/block-kit-builder

santam85:
This refers to the disclaimer at the top of https://api.slack.com/docs/message-attachments
In short, Attachements might get deprecated shortly in favor of blocks, which offer a much larger set of option to format messages.

UncleFirefox:
I think I might have to move to https://github.com/seratch/jslack for my project if this does not change soon.

johnnyleitrim:
Hi @UncleFirefox - We're starting to add support for Block Kit to our client at the moment.  You can try out the current development version by checking out the `block-kit` branch and building the library.

Most of the `...MessageParams` classes now have a `setBlocks(List<Block>)` method that you can use to add blocks to your messages

**Please note that this is still a work in progress, so the block kit elements may change or be removed.**

johnnyleitrim:
Version 1.4 has been released, which contains Block Kit support.

PR #110, opened 2019-11-11
#89: Add support for Block Kit

johnnyleitrim:
Adds support for [Block Kit](https://api.slack.com/reference/block-kit/blocks) when sending messages to slack.

Add example `PostABlockKitMessage` client is added as an example of how to send messages with block components.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
@@ -1,7 +1,10 @@
 package com.hubspot.slack.client.methods.params.chat;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+
+import org.immutables.value.Value;
 
 import com.hubspot.slack.client.methods.interceptor.HasChannel;
 import com.hubspot.slack.client.models.Attachment;
@@ -12,5 +15,8 @@ public interface MessageParams extends HasChannel {
 
   List<Attachment> getAttachments();
 
-  List<Block> getBlocks();
+  @Value.Default
+  default List<Block> getBlocks() {
+    return Collections.emptyList();
+  }
 }
--- a/slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
@@ -5,9 +5,12 @@ import java.util.Optional;
 
 import com.hubspot.slack.client.methods.interceptor.HasChannel;
 import com.hubspot.slack.client.models.Attachment;
+import com.hubspot.slack.client.models.blocks.Block;
 
 public interface MessageParams extends HasChannel {
   Optional<String> getText();
 
   List<Attachment> getAttachments();
+
+  List<Block> getBlocks();
 }
--- a/slack-base/src/main/java/com/hubspot/slack/client/methods/params/reactions/ReactionsAddParamsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/methods/params/reactions/ReactionsAddParamsIF.java
@@ -17,7 +17,21 @@ public interface ReactionsAddParamsIF {
   String getName();
   Optional<String> getChannel();
   Optional<String> getTimestamp();
+  /**
+   * @deprecated As of 22nd August 2019, Slack has removed support for file comments -
+   * <a href="https://api.slack.com/changelog/2018-05-file-threads-soon-tread#whats_changed">
+   *   file threads are the replacement
+   * </a>
+   */
+  @Deprecated
   Optional<String> getFile();
+  /**
+   * @deprecated As of 22nd August 2019, Slack has removed support for file comments -
+   * <a href="https://api.slack.com/changelog/2018-05-file-threads-soon-tread#whats_changed">
+   *   file threads are the replacement
+   * </a>
+   */
+  @Deprecated
   Optional<String> getFileComment();
 
   @Value.Check
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
@@ -22,5 +22,6 @@ public interface ActionsIF extends Block {
     return TYPE;
   }
 
+  @Value.Parameter
   List<BlockElement> getElements();
 }
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
@@ -17,7 +17,7 @@ public interface ActionsIF extends Block {
   String TYPE = "actions";
 
   @Override
-  @Value.Default
+  @Value.Derived
   default String getType() {
     return TYPE;
   }
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
@@ -1,0 +1,26 @@
+package com.hubspot.slack.client.models.blocks;
+
+import java.util.List;
+
+import org.immutables.value.Value;
+import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.blocks.elements.BlockElement;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ActionsIF extends Block {
+  String TYPE = "actions";
+
+  @Override
+  @Value.Default
+  default String getType() {
+    return TYPE;
+  }
+
+  List<BlockElement> getElements();
+}
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContextIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContextIF.java
@@ -19,7 +19,7 @@ public interface ContextIF extends Block {
   String TYPE = "context";
 
   @Override
-  @Value.Default
+  @Value.Derived
   default String getType() {
     return TYPE;
   }
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContextIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContextIF.java
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.hubspot.immutables.style.HubSpotStyle;
 import com.hubspot.slack.client.models.blocks.json.ImageOrTextDeserializer;
+import com.hubspot.slack.client.models.blocks.objects.ImageOrText;
 
 @Immutable
 @HubSpotStyle
@@ -24,5 +25,5 @@ public interface ContextIF extends Block {
   }
 
   @JsonDeserialize(contentUsing = ImageOrTextDeserializer.class)
-  List<Object> getElements();
+  List<ImageOrText> getElements();
 }
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContextIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContextIF.java
@@ -1,0 +1,26 @@
+package com.hubspot.slack.client.models.blocks;
+
+import java.util.List;
+
+import org.immutables.value.Value;
+import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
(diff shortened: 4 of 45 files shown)


johnnyleitrim:
Thanks @szabowexler for the feedback.  I should have addresses the issues you raised.  Can you have another look through and see if there's anything else that could be made clearer?

slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
```diff
@@ -5,9 +5,12 @@
...
 import com.hubspot.slack.client.methods.interceptor.HasChannel;
 import com.hubspot.slack.client.models.Attachment;
+import com.hubspot.slack.client.models.blocks.Block;
 
 public interface MessageParams extends HasChannel {
   Optional<String> getText();
 
   List<Attachment> getAttachments();
+
+  List<Block> getBlocks();
```
kecarroll: Will this not break implementations of this public interface?

slack-base/src/main/java/com/hubspot/slack/client/methods/params/reactions/ReactionsAddParamsIF.java
```diff
@@ -17,7 +17,21 @@
   String getName();
   Optional<String> getChannel();
   Optional<String> getTimestamp();
+  /**
+   * @deprecated As of 22nd August 2019, Slack has removed support for file comments -
+   * <a href="https://api.slack.com/changelog/2018-05-file-threads-soon-tread#whats_changed">
+   *   file threads are the replacement
+   * </a>
+   */
+  @Deprecated
```
kecarroll: The deprecation comment is identical to the file comments one. Should this not read `As of 22nd August 2019, Slack has removed support for files` ?

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
```diff
@@ -0,0 +1,26 @@
...
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.blocks.elements.BlockElement;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ActionsIF extends Block {
+  String TYPE = "actions";
```
kecarroll: Would this not be better as an enum?

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
```diff
@@ -0,0 +1,26 @@
...
+
+  @Override
+  @Value.Default
+  default String getType() {
```
kecarroll: Is the intention that the `type` can be manually set if needed? If not `@Derived` might be a better fit: https://immutables.github.io/immutable.html#derived-attributes

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
```diff
@@ -0,0 +1,26 @@
...
+    return TYPE;
+  }
+
+  List<BlockElement> getElements();
```
kecarroll: Definitely `List` and not `Set`?

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/InputIF.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  Text getLabel();
+
+  BlockElement getElement();
+
+  Optional<Text> getHint();
+
+  @JsonProperty("optional")
+  Optional<Boolean> isOptional();
```
kecarroll: An `Optional<Boolean>` to tell if it's optional?!

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
johnnyleitrim: Perfect!  That's exactly what I was looking for (but couldn't find it!)

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
johnnyleitrim: No - it's acceptable that you may have exactly the same block multiple times (there's no unique id, and you may want to show exactly the same block later in the message).

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/InputIF.java
johnnyleitrim: It's optionally optional!

slack-base/src/main/java/com/hubspot/slack/client/methods/params/reactions/ReactionsAddParamsIF.java
johnnyleitrim: The `file` field here was related to rendering file comments in the thread.  Slack aren't removing support for files, just file comments.  That's why the javadoc for both these fields reads the same.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
johnnyleitrim: I did think about that, but there was more code involved (since we need to send it as a string), and since this isn't a value that anyone should be interacting with, I decided it was easier to leave it as a `String` in each interface.  Do you think there would be an advantage to using an enum instead?

slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
johnnyleitrim: The only implementations we have of this interface are the immutable POJOs (which will be regenerated).  I wasn't able to find anyone directly extending this interface on github.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
kecarroll: I just thought it would be a better fit but your reasoning makes sense

slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
kecarroll: Fair enough so. I was thinking it might be safer to define it as

```java
default  List<Block> getBlocks() {
 return Collections.emptyList();
}
```

szabowexler: I don't think that's necessary. `List` isn't a breaking change on an immutable (it defaults to the empty list), so this method is actually unnecessary and just clutters the definition

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
szabowexler: Also I suspect you want these in order, right? Since set makes no guarantees that seems like it would be risky

slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
```diff
@@ -1,13 +1,22 @@
...
 import com.hubspot.slack.client.models.Attachment;
+import com.hubspot.slack.client.models.blocks.Block;
 
 public interface MessageParams extends HasChannel {
   Optional<String> getText();
 
   List<Attachment> getAttachments();
+
+  @Value.Default
+  default List<Block> getBlocks() {
```
szabowexler: Per my comment, you don't need this. Immutables default collections to the empty collection if they're not set, so you actually get this behavior without needing the explicit description of it here.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/DatePickerIF.java
```diff
@@ -0,0 +1,38 @@
...
+  @Value.Derived
+  default String getType() {
+    return TYPE;
+  }
+
+  String getActionId();
+
+  Optional<Text> getPlaceholder();
+
+  @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd")
```
szabowexler: TIL! This is really great, love this

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/UsersMultiSelectMenuIF.java
```diff
@@ -0,0 +1,37 @@
...
+  default String getType() {
+    return TYPE;
+  }
+
+  Text getPlaceholder();
+
+  String getActionId();
+
+  @JsonProperty("initial_users")
+  List<String> getInitialUserIds();
```
szabowexler: This feels like it should be a `Set`

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
```diff
@@ -0,0 +1,26 @@
...
+public interface ActionsIF extends Block {
+  String TYPE = "actions";
+
+  @Override
+  @Value.Derived
+  default String getType() {
+    return TYPE;
+  }
+
+  List<BlockElement> getElements();
```
szabowexler: Since this is the singleton field that folks need to set, I think we want to add `@Paramater` so that Immutables generates `Actions.of(...)` so you can just create such a block from a list of elements.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContextIF.java
```diff
@@ -0,0 +1,29 @@
...
+  String TYPE = "context";
+
+  @Override
+  @Value.Derived
+  default String getType() {
+    return TYPE;
+  }
+
+  @JsonDeserialize(contentUsing = ImageBlockOrTextDeserializer.class)
+  List<ImageBlockOrText> getElements();
```
szabowexler: Add `@Paramater`

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/SectionIF.java
```diff
@@ -0,0 +1,32 @@
...
+public interface SectionIF extends Block {
+  String TYPE = "section";
+
+  @Override
+  @Value.Derived
+  default String getType() {
+    return TYPE;
+  }
+
+  Text getText();
```
szabowexler: Add `@Paramater`

slack-java-client-examples/src/main/java/com/hubspot/slack/client/examples/Blocks.java
```diff
@@ -0,0 +1,72 @@
...
+        .build();
+  }
+
+  public static Text plainText(String text) {
+    return Text.builder()
+        .setText(text)
+        .setType(TextType.PLAIN_TEXT)
+        .build();
+  }
+
```
szabowexler: nit: extra new line

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
johnnyleitrim: Oh, I didn't know this was a thing!!  I'll go through each Immutable I've added and add `@Parameter` to all the required fields.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/UsersMultiSelectMenuIF.java
johnnyleitrim: Duplicates are allowed by Slack in this field, so I don't think `Set` would be appropriate.

johnnyleitrim: All the Johanns!

slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
johnnyleitrim: Keith has suggested adding this in would protect against anyone that might (who knows!) be implementing or extending `MessageParams` themselves.  I know it's not necessary for the immutables, but doing it this way makes this a non-breaking change.  What are your thoughts on this @szabowexler ?

szabowexler: Oh I see what you're saying. I think that's a violation of the contract here...this is really just intended as a way to make our annotation processor handle it. I don't really think we should be allowing people to extend it in arbitrary ways

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/UsersMultiSelectMenuIF.java
szabowexler: loooooool ok I stand corrected

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ActionsIF.java
szabowexler: (I've always used this in cases where there's a single field that can be used to construct it, so that you can build it with an `of` call)

szabowexler:
Let's ship it! Would you mind also cutting a release? Support for block kit is pretty sweet :)

slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/MessageParams.java
kecarroll: It's a public interface though. If it was just meant for internal immutables/annotation processing only then it probably shouldn't have been made public. As a safety measure I think the default implementation is the only way to not piss off users of this library

PR #114, opened 2019-11-20
User profile fields

jonathanwgoodwin:
These don't appear to be documented but do appear to exist. AFAICT its a `Map<String, ProfileField>`.

For  us for example we've got the actual username of the user in here, as well as other visible / non-visible custom profile data. 

Example redacted data:
```json
{
  "fields": {
    "X000000000": {
       "value": "",
       "alt": ""
     },
  }
}
```

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/users/ProfileFieldIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/users/ProfileFieldIF.java
@@ -1,5 +1,7 @@
 package com.hubspot.slack.client.models.users;
 
+import java.util.Optional;
+
 import org.immutables.value.Value.Immutable;
 
 import com.hubspot.immutables.style.HubSpotStyle;
@@ -10,5 +12,5 @@ public interface ProfileFieldIF {
 
   String getValue();
 
-  String getAlt();
+  Optional<String> getAlt();
 }
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/users/ProfileFieldIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/users/ProfileFieldIF.java
@@ -1,0 +1,14 @@
+package com.hubspot.slack.client.models.users;
+
+import org.immutables.value.Value.Immutable;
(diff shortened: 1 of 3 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/users/UserProfileIF.java
```diff
@@ -28,6 +29,9 @@
   Optional<String> getPhone();
   Optional<String> getSkype();
 
+  // Extra custom fields set by your workspace admin
+  Optional<Map<String, ProfileField>> getFields();
```
darcatron: Would be awesome if we could re-use [this deserialization test](https://github.com/HubSpot/slack-client/blob/master/slack-base/src/test/java/com/hubspot/slack/client/models/users/SlackUserDeserializationTest.java) to verify these fields are parsed

slack-base/src/main/java/com/hubspot/slack/client/models/users/ProfileFieldIF.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.hubspot.immutables.style.HubSpotStyle;
+
+@Immutable
+@HubSpotStyle
+public interface ProfileFieldIF {
+
+  String getValue();
+
+  String getAlt();
```
johnnyleitrim: Sounds like the `alt` field is [optional](https://api.slack.com/methods/users.profile.set):
> The `value` of a field is what should be displayed, unless the `alt` key is also present, in which case that is displayed instead.

szabowexler:
@jonathanwgoodwin what's the story here? Is this tested/ok to merge/still necessary? I'd like to either get it merged in or closed.

jonathanwgoodwin:
Turns out that this isn’t sent in the data for the list users call, only when you get an individual user account. That made the thing I wanted to use this for too slow so I’ll close this for now.

jonathanwgoodwin:
I'm back  🤣, the original use-case for this field didn't work out, but I've got a new place where I need to access these profile fields.

PR #117, opened 2019-12-06
Add default response type of in_channel.

gdaunton:
This adds a default response type of `in_channel` to `ChatPostMessageParams`.
This is an important field to include when responding in the channel to slash commands.

Currently, because no response type is supplied, responding to slash commands does not append the user's original command to the channel:
Screen Shot 2019-12-06 at 2 46 52 PM

This is a problem because other users looking at the channel will not be able to see what prompted the bot to output a message.

By providing the default of `in_channel` slack will now append the slash command executed by the user to the channel:
Screen Shot 2019-12-06 at 2 44 14 PM


gdaunton:
this actually doesn't seem to work as expected, closing for now.

gdaunton:
this is actually still necessary to be able to respond with ephemeral messages to slash command requests.

slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/AbstractChatPostMessageParams.java
```diff
@@ -11,13 +14,15 @@
...
 import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.TopLevelMessageResponseType;
 
 @Immutable
 @HubSpotStyle
 @JsonNaming(SnakeCaseStrategy.class)
 public abstract class AbstractChatPostMessageParams implements MessageParams {
   @JsonProperty("channel")
   public abstract String getChannelId();
+  public @Default TopLevelMessageResponseType getResponseType() { return IN_CHANNEL; }
```
szabowexler: Just a stylistic thing, can we make this multiline?

```java
@Default
public TopLevelMessageResponseType getResponseType() {
  return IN_CHANNEL;
}
```

gdaunton:
sure thing, so this is around sending an immediate response to a slash command request. If you don't send an immediate response, slack does not allow you to append the original command text to the channel which causes the behavior outlined in the top post. 

Here are the docs on sending an immediate response: https://api.slack.com/interactivity/slash-commands#responding_immediate_response

If the response_type is not set, it defaults to `ephemeral`:
> The `response_type` parameter in the JSON payload controls this visibility, by default it is set to `ephemeral`, but you can specify a value of `in_channel` to post the response into the channel, like this:

I've updated `ChatPostMessageParams` because slash command responses are supposed to match the structure of any other message that we are posting with the client:
>As mentioned, you can include more substantive info in the body of your `HTTP 200` response. In fact, you can use any of the complex formatting or Block Kit layout options that are available when sending any message.

szabowexler:
Is there a reason we are changing the default? It feels like this should just be an optional param (such that you can configure it if you want), but if Slack's default is ephemeral, why should our default be different?

gdaunton:
the default is specific to slash commands if the `response_type` value is not provided. Changing the default here to `ephemeral` would make all `ChatPostMessageParams` (used outside slash commands) `ephemeral` which would cause some problems. Should we make a slash command specific response object that has the default as `ephemeral`?

szabowexler:
What does this field mean outside of the context of a slash command? I assume since it's set when sending normal messages it can do something there too, right?

gdaunton:
ah yeah, looking at the docs regular messages don't have a `response_type` field: https://api.slack.com/methods/chat.postMessage. To send normal ephemeral messages we have  [`ChatPostEphemeralMessageParamsIF`](https://github.com/HubSpot/slack-client/blob/master/slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/ChatPostEphemeralMessageParamsIF.java).

gdaunton:
@szabowexler updated to make a new `SlashCommandResponse` object that is extended from a new base message class `AbstractChatMessageParams`. Unsure about the naming, is the convention still to name non-immutable abstract classes with `Abstract`?

gdaunton:
👍 I've tested it locally with no problems

gdaunton:
working on a branch deploy as well!

Issue #113, opened 2019-11-20
Support multipart file uploads


szabowexler:
https://github.com/HubSpot/slack-client/issues/107 revealed that we don't support the full `files.upload` API. This looks like a relatively simple extension to what already exists. The [files.upload](https://api.slack.com/methods/files.upload) documentation makes clear that there are two mechanisms for file upload:

1. Via the `content` parameter, which is what is currently supported (and is described in the docs as recommended "...for creating a "file" from a long message/paste and forces "editable" mode.", _and_
2. Via a `multipart/form-data` encoded `file` 

I think we would just need to add support for the second variety (which is effectively purely additive here).

Issue #107, opened 2019-10-07
Usage of file uploads


somejavadev:
Hi,  
  
I am trying to do a normal file upload but I cannot find any documentation on how to use the file upload functionality. I have tried the below but I think my problem comes in with the `content` of the builder:    
```java
 Result<FilesUploadResponse, SlackError> join = slackClient.uploadFile(FilesUploadParams.builder().setFilename("Test.png").setTitle("Test File").setContent("/tmp/test.png").build()).join();
```  
Do you by any chance have an example on how to input/format the content? I have tried Base64 encoding it as well without any luck. Ideally I would like to take a byte array and upload it.

johnnyleitrim:
Hi @somejavadev!

First off, you need to tell the slack client which channel(s) you want to post the file to (use `FilesUploadParams.setChannels()`)

Also, the `content` field is for the _actual raw content_ of the file, not the location of the file on your computer.

Below is an example that will post a text file to the slack channel referenced by the variable `channelToPostIn`:

```java
Result<FilesUploadResponse, SlackError> result = slackClient.uploadFile(FilesUploadParams.builder()
        .setChannels(Collections.singletonList(channelToPostIn))
        .setFilename("test.txt")
        .setTitle("Test File")
        .setContent("This is some plain text content")
        .build()).join();
```

noaHson86:
to be honest, `uploadFile()` doesnt provide the functionality you might expect.
providing the file as `String` is highly uncommon. uploading files(like .png or .xlsx) via this function will result in an unusable preview and files without further explanation.

somejavadev:
Thanks @johnnyleitrim as @noaHson86 pointed out the current file upload functionality in the library only supports raw text which is not what we want. It is pretty easy to just implement the slack API directly and not use this API as it doesn't provide all the functionality we required.

szabowexler:
This looks like a relatively simple extension to what already exists. The [files.upload](https://api.slack.com/methods/files.upload) documentation makes clear that there are two mechanisms for file upload:

1. Via the `content` parameter, which is what is currently supported (and is described in the docs as recommended "...for creating a "file" from a long message/paste and forces "editable" mode.", _and_
2. Via a `multipart/form-data` encoded `file` 

I think we would just need to add support for the second variety (which is effectively purely additive here).

PR #118, opened 2019-12-09
Add support for multipart upload of files

szabowexler:
This PR builds on multipart request support in https://github.com/HubSpot/Horizon/pull/30 to allow uploading files, and resolves https://github.com/HubSpot/slack-client/issues/113 and https://github.com/HubSpot/slack-client/issues/107. 

*Update*: we've elected to move the multipart processing directly into the slack client to avoid a new dep (and upgrade headache) in Horizon. The multipart http pieces are self-contained, so we can just do it here directly.


szabowexler:
I've tested this out with some basic PNG uploads. Good to note that slack has a bug right now that means if you upload a file (even to a public channel) and then try and use an `Image` block, they'll error when trying to fetch the "private" link. At present the `Image` block element requires a publicly accessible image to work.

szabowexler:
Going to merge this tomorrow morning.

PR #119, opened 2019-12-10
Add support for Views

### LINKED ISSUE #115: Add Modal support
Slack have [added Modal as the replacement for dialogs](https://api.slack.com/block-kit/dialogs-to-modals).

Modals have the ability to use Block Kit interactive components, and since this library now supports Block Kit, it would be useful to add Modal support as well.

For more information on using Modals, see [Reference: View payloads](https://api.slack.com/reference/surfaces/views)
- [x] Add POJOs for the fields specified at https://api.slack.com/reference/surfaces/views
  There should be different POJOs for Home and Modal, so that the relevant methods only get the correct amount of information.
- [x] add support for methods: 
  - views.open
  - views.update
  - views.push
  - views.publish

johnnyleitrim:
1. The main headline of this PR is that it adds support for views for Modal and Home surfaces.  It adds support for the following slack commands:

- `views.open`
- `views.update`
- `views.push`
- `views.publish`

2. We also add support for [Interactive components payloads](https://api.slack.com/reference/interaction-payloads/block-actions) - this allows the slack client to deserialise interactive messages that are received from block-kit components

3. We also add a simple SlackServer, which can be used during development of this client to test our deserialisation of messages received from Slack directly.

**There is one backward incompatible change in this PR**
The `channel` is now optional in `SlackInteractiveCallback`, since it is not included in all messages received from slack.  I'm not sure if there's a better way to make this change that makes it backward compatible.

_Please note that some of the POJO hierarchy decisions may seems strange, but there are a large number of restrictions when using Immutables, Jackson and polymorphic subtypes that needs to be included as JSON values_.

Closes #115


slack-base/src/main/java/com/hubspot/slack/client/models/interaction/SlackInteractiveCallback.java
```diff
@@ -11,18 +13,15 @@
...
   InteractiveCallbackType getType();
-
-  String getCallbackId();
-  String getActionTs();
   String getToken();
-
   SlackTeam getTeam();
-  SlackChannel getChannel();
   SlackUserLite getUser();
+  Optional<SlackChannel> getChannel();
```
szabowexler: Oof I see what you mean. Logistically it feels like this method should have lived on the discrete subclasses that actually have it (i.e. a `MessageAction` should still have a non-optional channel field). 

I'm really sad about breaking the API here, but I do recognize the reason is a mistake I made initially including this. @jhaber do you have any advice on the cleanest way to recover from an API level mistake like this? This field shouldn't have been on the base class, but now it's out in the wild.

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/json/BlockElementActionDeserializer.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.ObjectCodec;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.hubspot.slack.client.models.blocks.elements.BlockElement;
+import com.hubspot.slack.client.models.interaction.BlockElementAction;
+
+public class BlockElementActionDeserializer extends StdDeserializer<BlockElementAction> {
```
szabowexler: Just a quick q: can you comment on why we needed to go the explicit deserializer route rather than straight jackson?

slack-java-client-examples/src/main/java/com/hubspot/slack/server/example/SlackMessageHandler.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(SlackMessageHandler.class);
+
+  private final SlackClient slackClient;
+
+  public SlackMessageHandler() {
+    slackClient = BasicRuntimeConfig.getClient();
+  }
+
+
```
szabowexler: nit: extra newline

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/json/BlockElementActionDeserializer.java
johnnyleitrim: The problem is that you can't deserialise a polymorphic type with `@JsonUnwrapped` using Jackson. Have a look at [this stackoverflow question](https://stackoverflow.com/questions/37423848/deserializing-polymorphic-types-with-jsonunwrapped-using-jackson) for a summary

There was a bug raised about this against jackson years ago, but was closed as [will not fix](https://markmail.org/message/cyeyc2ousjp72lh3)

There's a similar issue where the maintainers also said it would not be supported. https://github.com/FasterXML/jackson-databind/issues/81

I spent a ridiculous amount of time on this (since I _hate_ having to write a customer serialiser/deserialiser), but I couldn't figure out any other way to achieve t.

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/SlackInteractiveCallback.java
szabowexler: Chatted a bit with @stevegutz and @wsorenson about the implications of this change. This shouldn't have been done as it was, and it's not an API change but rather a build failure (and one that shouldn't be too hard to correct). In light of that, we're going to just roll ahead (since it will even be ok in many situations at present since folks don't extend this class, generally).

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/json/BlockElementActionDeserializer.java
szabowexler: 👍 thanks for the explanation

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/SlackInteractiveCallback.java
stevegutz: Can you explain what you mean by "build failure"? A build failure where? This change can break (at runtime) any code that extends the base class. I don't think we need to worry about that, but I want to make sure this is documented in the PR.

wsorenson: I didn't look closely at the change, but I suppose it brings into question if any of these options are available:

1) Only use explicit library versions in Maven.
2) Provide a second hierarchy of models and deprecate the original
3) Rename this class

Do any of these make sense?

johnnyleitrim: I'm in favour of (1), since that's kinda what versions are for.  2 & 3 are tricky because these classes are used for deserialisation of the messages received from slack.

PR #123, opened 2019-12-12
Handle unknown block and block element types

johnnyleitrim:
Handle unknown block and block element types by having a default Unknown type for each.

On [this page](https://api.slack.com/changelog/2019-09-what-they-see-is-what-you-get-and-more-and-less), Slack have said:

> Expect to encounter other (even undocumented) elements.

Therefore, we will likely never know beforehand the entire list of blocks and block elements that may be returned, so this is our mechanism for handling unknown types.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/UnknownBlock.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/UnknownBlock.java
@@ -1,0 +1,18 @@
+package com.hubspot.slack.client.models.blocks;
+
+import java.util.Optional;
+
+public class UnknownBlock implements Block {
+  protected UnknownBlock() {
+  }
+
+  @Override
+  public String getType() {
+    return "";
+  }
+
+  @Override
+  public Optional<String> getBlockId() {
+    return Optional.empty();
+  }
+}
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/UnknownBlock.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/UnknownBlock.java
@@ -3,12 +3,14 @@ package com.hubspot.slack.client.models.blocks;
 import java.util.Optional;
 
 public class UnknownBlock implements Block {
+  public static final String TYPE = "unknown";
+
   protected UnknownBlock() {
   }
 
   @Override
   public String getType() {
-    return "";
+    return TYPE;
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/blocks/UnknownBlock.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Optional;
+
+public class UnknownBlock implements Block {
+  protected UnknownBlock() {
+  }
+
+  @Override
+  public String getType() {
+    return "";
```
szabowexler: is there any way we can make the unknown block actually pick up whatever type they sent back to us, _or_ `UNKNOWN` if there's nothing? I'd love to be able to inspect this in some cases and actually see what type we're missing.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/UnknownBlockElement.java
```diff
@@ -0,0 +1,11 @@
+package com.hubspot.slack.client.models.blocks.elements;
+
+public class UnknownBlockElement implements BlockElement {
+  protected UnknownBlockElement() {
+  }
+
+  @Override
+  public String getType() {
+    return "";
```
szabowexler: Same as above, I'd really like to see this actually inherit whatever type we get back, when we deserialize things using this

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/UnknownBlock.java
johnnyleitrim: I didn't want to go with UNKNOWN (or any concrete value) incase it ever clashes with one of slack's actual blocks (since they have undocumented ones).

I'll investigate capturing the actual value returned from slack though.

johnnyleitrim: Jackson doesn't support capturing the original type value of an unknown value.  I'll need to write customer serialiser and deserialisers for that.

szabowexler: I think this is a case where that would be worthwhile, since having to manually debug and figure out the raw strings is going to be pretty painful

johnnyleitrim: There's no way to write the deserialiser without having to hand code the deserialisation of every field in every block/block element type.  For now, let's go with `unknown` as the type.  If someone needs support for a new type of block or element, they can always add it.

szabowexler: Ok I buy that. Let's do it

PR #129, opened 2020-01-01
app_home_opened event support

lightbody:
This little patch moves the `app_home_opened` incoming event from the realm of being unknown (`SlackEventType.UNKNOWN`) and thus turned into a simple `SlackEventSkeleton` object to something more useful: a full blown `SlackAppHomeOpenedEvent` that contains the current view payload.

PS: Not sure if the package name "bot" or "views" would have been better.


lightbody:
Now that I have several PRs, I'm breaking them up into branches. I'm closing this one and going to re-submit from a separate branch I'm maintaining.

Issue #128, opened 2019-12-31
Cannot parse JSON into SlackEventWrapper for many events


lightbody:
This is similar to issue #52, which was resolved but seems to still be a problem for me. Specifically, I am trying to parse an incoming event action like so:

```java
objectMapper.readValue(request.getBody(String.class).get(), SlackEventWrapper.class)
```

For some events, like a message, it works wonderfully. But for others, where there is no `ts` field, I get this error from Jackson:

```
Cannot construct instance of `com.hubspot.slack.client.models.events.SlackEventSkeleton`, problem: Cannot build SlackEventSkeleton, some of required attributes are not set [ts]
```
I would love to use all the event models that are built up in this project, as I'm already using the web API and it's working great, but this is a bit of a showstopper for me. Any ideas how to work around it?

Thank you for the great project!

szabowexler:
I suspect you may have fixed this for yourself :)

If that's the case, want to close this issue out?

If not, I think the answer here is that we need to adjust the class hierarchy to ensure that things that don't have a referenced message (like e.g. an app event) don't inherit the exact same fields via the same class hierarchy as message events.

lightbody:
Just a quick update on this: I fixed this specifically for the `app_home_opened` event... by creating a `SlackAppHomeOpenedEvent` event (issue #129). But more generally, my code simply can never parse the event wrappers into`SlackEventWrapper` if there isn't a specific event class mapped in `SlackEventType` enum.

I agree that there is something wrong with the class hierarchy. There have been a number of issues along the way related to `ts`, which seems to be the offended issue. But the fix to this is a bit out of my reach out at the moment. Either way, I need new event classes for `app_uninstalled`, `tokens_revoked`, and `dnd_updated` so I'm going to keep working around this the same way I did for `app_home_opened`. The only weird thing is that `app_uninstalled` has no body, so my patch for that one probably should get some extra scrutiny :)

PR #131, opened 2020-01-02
app_home_opened event support

lightbody:
This little patch moves the app_home_opened incoming event from the realm of being unknown (SlackEventType.UNKNOWN) and thus turned into a simple SlackEventSkeleton object to something more useful: a full blown SlackAppHomeOpenedEvent that contains the current view payload.

PS: Not sure if the package name "bot" or "views" would have been better.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/events/bot/SlackAppHomeOpenedEventIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/events/bot/SlackAppHomeOpenedEventIF.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.methods.interceptor.HasUser;
 import com.hubspot.slack.client.models.events.SlackEvent;
 import com.hubspot.slack.client.models.response.views.HomeTabViewResponseIF;
 import org.immutables.value.Value;
@@ -15,7 +16,7 @@ import java.util.Optional;
 @HubSpotStyle
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
 @JsonDeserialize(as = SlackAppHomeOpenedEvent.class)
-public interface SlackAppHomeOpenedEventIF extends SlackEvent {
+public interface SlackAppHomeOpenedEventIF extends SlackEvent, HasUser {
     String getTab();
 
     Optional<HomeTabViewResponseIF> getView();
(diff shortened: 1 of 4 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/events/bot/SlackAppHomeOpenedEventIF.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Optional;
+
+@Value.Immutable
+@HubSpotStyle
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+@JsonDeserialize(as = SlackAppHomeOpenedEvent.class)
+public interface SlackAppHomeOpenedEventIF extends SlackEvent, HasUser {
+    String getTab();
+
+    Optional<HomeTabViewResponseIF> getView();
```
johnnyleitrim: You should reference the object type `HomeTabViewResponse` rather than the interface.

slack-base/src/main/java/com/hubspot/slack/client/models/events/bot/SlackAppHomeOpenedEventIF.java
```diff
@@ -0,0 +1,36 @@
...
+
+    Optional<HomeTabViewResponse> getView();
+
+    @JsonProperty("user")
+    String getUserId();
+
+    @JsonProperty("channel")
+    String getChannelId();
+
+    //Home opened events do not have a ts, so we manually set it as null
```
szabowexler: This makes me nervous, since it violates our implicit invariant that any nullable field is wrapped in`Optional` and any non-`Optional` field is not `null`. What it suggests to me is that our initial class hierarchy was not correct, and that app events (i.e. things that may lack a `ts` field`) belong in a slightly different inheritance tree.

I haven't dug deeply into the difference in shapes, but I could imagine having a base `SlackEvent` that contains all common fields, and then another `HasReferencedMessage` interface that adds the `ts` field. We could add an intermediate `SlackMessageEvent`to capture slack events which also have a `ts` field, and then we wouldn't need to do this hack.

Does that make sense?

lightbody: Makes total sense to me. I was just following the pattern that was already there for some other events. I'll see if I can shuffle things around like you suggest, but it might be a few days and it also might be little out of my comfort zone, though I'm starting to get more and more familiar with this codebase :)

szabowexler: If this is how things already are and this is blocking you, we can go ahead and copy the existing pattern to get you moving. If you don't mind taking a stab at this style of cleanup, if you find it's not too onerous we'd appreciate it! :)

Thanks!

lightbody: See my comment in issue #128 re the class cleanup. I'm already on my way (I have a fork I'm using in prod), so your call on whether you want to commit this or wait for the bigger cleanup on `ts`. I just don't know how much I can help on that front.

szabowexler:
sorry for delay on merge!

PR #133, opened 2020-01-04
bug fix: logic matches error text... and documented expectations :)

lightbody:


slack-base/src/main/java/com/hubspot/slack/client/methods/params/views/UpdateViewParamsIF.java
```diff
@@ -27,7 +27,9 @@
 
   @Check
   default void check() {
-    Preconditions.checkState(Strings.isNullOrEmpty(getExternalId().orElse(null)) && Strings.isNullOrEmpty(getViewId().orElse(null)),
+    final boolean externalIdSet = !Strings.isNullOrEmpty(getExternalId().orElse(null));
+    final boolean viewIdSet = !Strings.isNullOrEmpty(getViewId().orElse(null));
+    Preconditions.checkState((externalIdSet || viewIdSet) && !(externalIdSet && viewIdSet),
```
szabowexler: What about `Boolean#logicalXor()`? Seems to be what you're going for here.

szabowexler:
close enough we'll just merge it

PR #134, opened 2020-01-04
supported for selected_option as an alternative to the value field

lightbody:
Even though the Slack API docs say that when an user select's a menu, the interaction's value should arrive in the "value" field, I believe the docs are incorrect and reality is different.

Specifically, what I am seeing in the real world is that select menus will have their value arrive instead as a "selected_option" field. Thus, this deserializer will inspect for this field and pull the value from there if present.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/json/BlockElementActionDeserializer.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/json/BlockElementActionDeserializer.java
@@ -13,6 +13,7 @@ import com.hubspot.slack.client.models.interaction.BlockElementAction;
 
 public class BlockElementActionDeserializer extends StdDeserializer<BlockElementAction> {
   private static final String BLOCK_ID_FIELD = "block_id";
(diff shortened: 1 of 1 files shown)


lightbody:
I'm sure you'd like a test for this one... any existing tests I can look at for inspiration?

szabowexler:
Agreed on the testing front -- I'm not sure we have something exactly like this though. I'd peek through the existing block kit tests to see if there's anything close to what you're getting at here. I think basically the idea should be to capture a payload you get from slack that currently causes the client to complain, and then add that as a deserialization test, and just verify your changes make a previously impossible to deserialize payload (that you saw in the wild) actually pass.

szabowexler:
hey @lightbody -- wanted to circle back to this. What's our story here?

lightbody:
Apologies - been heads down gearing up for our big launch (https://lifeworkcalendar.com if you're curious) and then my 3yo son had surgery last week to top it off. I'll get back to this soon!

lightbody:
This was handled by PR #166, closing now.

PR #135, opened 2020-01-04
common interface for BlockElements that also have an action_id

lightbody:
Even though the Slack API docs say that when an user select's a menu, the interaction's value should arrive in the "value" field, I believe the docs are incorrect and reality is different.

Specifically, what I am seeing in the real world is that select menus will have their value arrive instead as a "selected_option" field. Thus, this deserializer will inspect for this field and pull the value from there if present.


szabowexler:
I think the text for this PR got confused -- it looks like it adds an "action id" interface wrapping (surprise!) an action id. Slack does reference [this field](https://api.slack.com/reference/block-kit/block-elements), so it seems like this was just an oversight. Thanks for the PR!

PR #137, opened 2020-01-06
removes logging for deprecated field that has since been removed

lightbody:
This removes the warnings about getReplies() being deprecated. I was finding that it was firing even thought I wasn't calling it, I believe because Jackson serialization was tickling the method. Rather than try to solve that issue, I realized that because the functionality has since been removed (October 2019), it probably makes sense to just delete it. 

If you agree, here is the PR.

If not, I'd like to figure out a way to get the warnings to stop :)


szabowexler:
Agreed, it's time. Thanks!

PR #145, opened 2020-01-29
turns out the tab field appears to be optional in the real world :(

lightbody:
Got this error in our logs today. After re-reading the docs, it looks like they don't clear say whether it should always be sent or not, but given that the real world seems to push me events without a tab field, I think you should consider this change:

https://api.slack.com/events/app_home_opened

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/events/bot/SlackAppHomeOpenedEventIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/events/bot/SlackAppHomeOpenedEventIF.java
@@ -7,7 +7,7 @@ import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.hubspot.immutables.style.HubSpotStyle;
 import com.hubspot.slack.client.methods.interceptor.HasUser;
 import com.hubspot.slack.client.models.events.SlackEvent;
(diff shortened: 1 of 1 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/events/bot/SlackAppHomeOpenedEventIF.java
```diff
@@ -17,9 +17,9 @@
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
 @JsonDeserialize(as = SlackAppHomeOpenedEvent.class)
 public interface SlackAppHomeOpenedEventIF extends SlackEvent, HasUser {
-    String getTab();
+    Optional<String> getTab();
 
-    Optional<HomeTabViewResponse> getView();
+    Optional<HomeTabViewResponseIF> getView();
```
szabowexler: We prefer to use the concrete instance of the class, rather than the interface (which just exists so our annotation processor knows to generate the desired POJO)

lightbody: Doh! That was already pointed out by you and corrected by me once before, I just messed up merging an older changeset in. Someday I'll get comfortable with git :) Thanks, I'll fix shortly.

PR #151, opened 2020-02-12
Implmented conversations.join

harmingcola:
Implemented `conversations.join` request.


slack-base/src/main/java/com/hubspot/slack/client/methods/params/conversations/ConversationsJoinParamsIF.java
```diff
@@ -0,0 +1,18 @@
...
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ConversationsJoinParamsIF {
+
+  /*
+   * ID of channel to join
+   */
+  String getChannel();
```
szabowexler: Let's name this `getChannelId` to make clear it's an ID (without needing the comment), and use `@JsonProperty` to rename it to what slack expects. In general we prefer to avoid needing comments, since they tend to indicate unclear/inadequate naming

PR #155, opened 2020-02-27
Add token to postMessage params

bdros:


slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/AbstractChatMessageParams.java
```diff
@@ -15,6 +15,7 @@
 public abstract class AbstractChatMessageParams implements MessageParams {
   @JsonProperty("channel")
   public abstract String getChannelId();
+  public abstract String getToken();
```
szabowexler, line 18: do you need this? I would have expected that given that this implements `HasToken` via `MessageParams`, you shouldn't need to explicitly define it here, right? Failing that, seems like it should have an `@Override` tag

PR #156, opened 2020-03-12
First pass @ supporting these payloads

wsorenson:


slack-base/src/main/java/com/hubspot/slack/client/models/interaction/ViewSubmissionIF.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ViewSubmissionIF extends SlackInteractiveCallback {
+
+  ViewInteractionPayload getView();
+
+  @Override
+  default SlackChannel getChannel() {
```
szabowexler: We should come back and fix the class hierarchy at some point, this is definitely going to burn people =/

PR #175, opened 2020-06-18
Add model for Slack member_left_channel event

zmarushchak-hs:
Need to deal with [member_left_channel](https://api.slack.com/events/member_left_channel) Slack event. Added model for it.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/events/user/SlackMemberLeftChannelEventIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/events/user/SlackMemberLeftChannelEventIF.java
@@ -1,0 +1,34 @@
+package com.hubspot.slack.client.models.events.user;
+
+import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.ChannelType;
+import com.hubspot.slack.client.models.events.SlackEvent;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonDeserialize(as = SlackMemberLeftChannelEvent.class)
+public interface SlackMemberLeftChannelEventIF extends SlackEvent {
+
+  @JsonProperty("user")
+  String getUserId();
+
+  @JsonProperty("channel")
+  String getChannelId();
+
+  ChannelType getChannelType();
+
+  String getTeam();
+
+  //Member joined events do not have a ts, so we manually set it as null
+  @Override
+  default String getTs() {
+    return null;
+  }
+}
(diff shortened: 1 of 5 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/events/user/SlackMemberLeftChannelEventIF.java
```diff
@@ -0,0 +1,34 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.ChannelType;
+import com.hubspot.slack.client.models.events.SlackEvent;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+@JsonDeserialize(as = SlackMemberLeftChannelEvent.class)
```
omotnyk, line 16: Why do we need this?

omotnyk: I mean, I see that it's copied from other events but still, why do we need this?

omotnyk: Okay, I see. Parent interface has custom deserializer.

omotnyk:
Wait, ignore my approval.

slack-base/src/test/resources/member_left_public_channel.json
```diff
@@ -0,0 +1,19 @@
...
+  "token": "redacted",
+  "team_id": "redacted",
+  "api_app_id": "redacted",
+  "event": {
+    "type": "member_left_channel",
+    "user": "redacted",
+    "channel": "redacted",
+    "channel_type": "C",
+    "team": "redacted",
+    "event_ts": "1540822163.000100"
```
omotnyk, line 11: I thought this event has no TS.

slack-base/src/test/resources/member_left_private_channel.json
```diff
@@ -0,0 +1,19 @@
...
+  "token": "redacted",
+  "team_id": "redacted",
+  "api_app_id": "redacted",
+  "event": {
+    "type": "member_left_channel",
+    "user": "redacted",
+    "channel": "redacted",
+    "channel_type": "G",
+    "team": "redacted",
+    "event_ts": "1540822163.000100"
```
omotnyk, line 11: I thought this event has no TS.

omotnyk: `//Member joined events do not have a ts, so we manually set it as null`

slack-base/src/test/resources/member_left_public_channel.json
omotnyk: `//Member joined events do not have a ts, so we manually set it as null`

omotnyk: Discussed offline.
So, this event has it but it isn't explicitly stated in the documentation and it doesn't seem like anybody needs that timestamp so we'll proceed with current implementation but will leave this test JSON with `event_ts` since we receive it sometimes.

slack-base/src/test/resources/member_left_private_channel.json
omotnyk: Same as above.

PR #193, opened 2020-09-14
Add Header Block implementation

### LINKED ISSUE #190: Add support for Header block type for message API
**Is your feature request related to a problem? Please describe.**

There is a block type which I don't believe is currently available in our hubspot slack-integration: [Header](https://api.slack.com/reference/block-kit/blocks#header). If I had to guess, I believe this is because the Header type is a relatively new type. I think it would be useful to add, as it formats a message like a markdown heading, making it easier to read longer messages.

**Describe alternatives you've considered**
I'm thinking about implementing a `Header` class directly by extending the `Section` block and setting its type to always be "Header"

richardxu100:
Closes #190 

I modeled this off of `SectionIF.java` and the Header API described [here](https://api.slack.com/reference/block-kit/blocks#header). I'm very new to java, so any comments or suggestions would be very appreciated.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/HeaderIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/HeaderIF.java
@@ -2,14 +2,11 @@ package com.hubspot.slack.client.models.blocks;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Strings;
 import com.hubspot.slack.client.models.blocks.objects.Text;
 import com.hubspot.slack.client.models.blocks.objects.TextType;
 import org.immutables.value.Value.Immutable;
 import com.hubspot.immutables.style.HubSpotStyle;
 import org.immutables.value.Value.Check;
-import org.immutables.value.Value.Default;
-import javax.annotation.Nullable;
 import org.immutables.value.Value;
 
 @Immutable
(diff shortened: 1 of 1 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/blocks/HeaderIF.java
```diff
@@ -0,0 +1,45 @@
...
+  @Value.Derived
+  default String getType() {
+    return TYPE;
+  }
+
+  @Default
+  @Nullable
+  @Value.Parameter
+  default Text getText() {
+    return null;
```
szabowexler, line 31: Given that your precondition check indicates the header can't have a null text block, what you want to do here is remove the entire default block, and make this just:

```java
@Value.Parameter
Text getText();
```

Then you can assume text is not null, which simplifies your check block below.

szabowexler (changes requested):

richardxu100:
Thanks for the suggestion @szabowexler! Please let me know if there's anything else that would be helpful to add to this PR

szabowexler:
Sure! The final check here is for you to confirm that you've posted a sample message, and it worked. Once you do that, just take a quick screenshot of the result and drop it here for posterity, and we'll go ahead and merge. You won't be able to use it immediately until we cut a release, but we can try and do that sometime towards the end of this week.

richardxu100:
> Sure! The final check here is for you to confirm that you've posted a sample message, and it worked. Once you do that, just take a quick screenshot of the result and drop it here for posterity, and we'll go ahead and merge. You won't be able to use it immediately until we cut a release, but we can try and do that sometime towards the end of this week.

Screen Shot 2020-09-15 at 10 16 10 AM

I was able to test that the header block is working properly! I'm not sure however if my method of testing is sufficient, as I basically just copied this file into my project that is using the `slack-client`. I think a more proper way to test this library would be to somehow import my local, modified `slack-client` into my project, instead of using the global version of this library, though I'm not sure how this would be done as I'm very new to java development. Do you have any suggestions on how to test this PR more properly, or is the testing I did sufficient?

PR #199, opened 2020-10-20
Add OAuth credentials object for v2 Slack app

zmarushchak-hs:
Object added in order to be able to deserialize response for [oauth.v2.access](https://api.slack.com/methods/oauth.v2.access) request which is used on Slack app installation to exchange temporary OAuth code for an API access token.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/auth/OAuthV2CredentialsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/auth/OAuthV2CredentialsIF.java
@@ -15,19 +15,19 @@ import com.hubspot.slack.client.models.users.SlackUserLite;
 @HubSpotStyle
 @JsonNaming(SnakeCaseStrategy.class)
 public abstract class OAuthV2CredentialsIF {
-  public abstract String getAccessToken();
-
-  @JsonProperty("team")
-  public abstract SlackTeamLite getSlackTeamLite();
+  @JsonProperty("access_token")
+  public abstract String getBotAccessToken();
+  public abstract SlackTeamLite getTeam();
   public abstract Optional<String> getScope();
   public abstract SlackUserLite getAuthedUser();
+  public abstract String getBotUserId();
 
   public String getTeamId() {
-    return getSlackTeamLite().getId();
+    return getTeam().getId();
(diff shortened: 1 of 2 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/auth/OAuthV2CredentialsIF.java
```diff
@@ -0,0 +1,36 @@
...
+import com.hubspot.slack.client.models.users.SlackUserLite;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public abstract class OAuthV2CredentialsIF {
+  public abstract String getAccessToken();
+
+  @JsonProperty("team")
+  public abstract SlackTeamLite getSlackTeamLite();
```
omotnyk, line 21: I think this can be just `getTeam`. This way we don't need to override json property name.

slack-base/src/main/java/com/hubspot/slack/client/models/auth/OAuthV2CredentialsIF.java
```diff
@@ -0,0 +1,36 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.teams.SlackTeamLite;
+import com.hubspot.slack.client.models.users.SlackUserLite;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public abstract class OAuthV2CredentialsIF {
+  public abstract String getAccessToken();
```
omotnyk, line 18: I wonder if we should rename this to `getBotAccessToken`.

omotnyk (commented):
We heavily rely on bot user ID. Should we store it? Also, it seems like this changes the auth process completely. As far as I see, https://api.slack.com/methods/oauth.v2.access still returns both user and bot token but this model doesn't store user token. Is this intentional? Will we ever need this token again? right now we do a lot of things using user token. Will the migration process seamless in this case?

omotnyk (approved):

PR #200, opened 2020-10-23
Don't include 'as_user' param in chat.postEphemeral method for V2 app

zmarushchak-hs:
V2 Slack app doesn't support `as_user` param for chat.postEphemeral method.  
Here is the error:
```
{
  "ok" : false,
  "error" : "invalid_arguments",
  "deprecated_argument" : "as_user",
  "warning" : "missing_charset",
  "response_metadata" : {
    "warnings" : [ "missing_charset" ]
  }
```

This PR adds methods to not include `as_user` param in these methods.

Please see also:
https://api.slack.com/authentication/quickstart#calling


omotnyk (changes requested):
I think we can annotate [this](https://github.com/HubSpot/slack-client/blob/8e09690ba2aba2c5eaee8ce7402e9f1ba872ab5c/slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/ChatPostEphemeralMessageParamsIF.java#L23) with `@JsonInclude(JsonInclude.Include.NON_ABSENT)` instead.

PR #202, opened 2020-10-27
Add models for group events

nliutyi:
As slack actually sends us a lot more events, then we have now, and to satisfy need of having consumed events from private channels (so-called groups in the Slack API), this PR contains a few added models, which are also binded with the SlackEventType enum entries.

PR is ready for review.


omotnyk (commented):
All the channel-related events override `getTs` because events actually have no TS but the field is mandatory in the immutable. Do we also need to do this with group events? Do they have TS?
```java
  @Override
  default String getTs() {
    return null;
  }
```

nliutyi:
@omotnyk I actually don't see it here [this](https://api.slack.com/events/group_unarchive) but it's not hard to add this stuff.
image

omotnyk (commented):
Let's add it then. I also think that we can have `SlackGroupEventCore` that contains `channel` instead of declaring the field in every immutable.

nliutyi:
@omotnyk Your comments are addressed.

omotnyk (approved):

PR #204, opened 2020-11-02
Update group and channel models. Add general SlackConversationEventCore model.

nliutyi:
While working with this repo, I've discovered, that there are actually no general parent class, which will combine channel and group events, besides SlackEvent. 
For purposes of better compatibility, SlackConversationEventCore has been introduced.
Also, adding SlackConversationEventWithChannel, for those events, which require usage of full SlackChannel object, not only the id.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/events/conversation/SlackConversationEventWithChannel.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/events/conversation/SlackConversationEventWithChannel.java
@@ -1,0 +1,15 @@
+package com.hubspot.slack.client.models.events.conversation;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.hubspot.slack.client.models.SlackChannel;
+import org.immutables.value.Value;
+
+public interface SlackConversationEventWithChannel extends SlackConversationEventCore {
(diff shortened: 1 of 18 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/events/conversation/SlackConversationEventWithChannel.java
```diff
@@ -0,0 +1,15 @@
...
+
+public interface SlackConversationEventWithChannel extends SlackConversationEventCore{
+    SlackChannel getChannel();
+
+    @JsonIgnore
+    @Value.Derived
+    default String getChannelId() {
+        return getChannel().getId();
+    }
+}
```
omotnyk, line 15: I think the formatting is a bit off in this file.

omotnyk (commented):
One nit.

omotnyk (commented):
Let's also add integration tests for group events to https://github.com/HubSpot/slack-client/blob/a867e519a0cf83cedb45eb5f5f2e295995b6798f/slack-base/src/test/java/com/hubspot/slack/client/models/events/json/EventDeserializerTest.java.

omotnyk (approved):

PR #207, opened 2020-11-20
implements users_profile_set + dnd methods

lightbody:
This PR implements the users_profile_set API. I will add in test cases as soon as I get a moment :)


szabowexler (approved):
seems legit to me

PR #216, opened 2021-03-19
Use conversations_replies instead of channels_replies

ovcica:
channels_replies has been deprecated as of Feb 24th: https://api.slack.com/changelog/2020-01-deprecating-antecedents-to-the-conversations-api

Conversations API doc: https://api.slack.com/methods/conversations.replies

It seems the response format is the same, so the same parsing should continue to work.

@szabowexler @kecarroll


ovcica:
We'll fix this in a different way within HubSpot

PR #217, opened 2021-03-24
Added support of channel_shared event

leonidhladii:
**Problem**
Missing [channel_shared](https://api.slack.com/events/channel_shared#:~:text=The%20channel_shared%20event%20is%20sent,to%20an%20existing%20shared%20channel.) event.

**Description**
Added support of channel shared event to `SlackEventType`


slack-base/src/main/java/com/hubspot/slack/client/models/events/channel/SlackChannelSharedEventIF.java
```diff
@@ -0,0 +1,18 @@
...
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.events.conversation.SlackConversationEventCore;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
```
zmarushchak-hs, line 13: Please correct me if I'm wrong, but I think we don't need `@JsonNaming(SnakeCaseStrategy.class)` if we have `@JsonProperty("connected_team_id")` set for the field.

slack-base/src/test/resources/channel_shared_event.json
```diff
@@ -0,0 +1,17 @@
...
+    "channel": "C06EQBRR6",
+    "event_ts": "1561064063.001100"
+  },
+  "type": "event_callback",
+  "event_id": "EvCYA72LVA",
+  "event_time": 1537542788,
+  "authed_users": [
+    "UBRFHDFGA"
+  ]
+}
```
zmarushchak-hs, line 17: New line, please.

zmarushchak-hs (commented):
Please see my comments inlined.

PR #222, opened 2021-04-01
send slack token properly

tpoll:
https://api.slack.com/changelog/2020-11-no-more-tokens-in-querystrings-for-newly-created-apps

New apps currently can't use these methods.

We need to start sending all the tokens as Auth headers and not query params.


tpoll:
Also -- reformats the imports to match our style guide.

PR #224, opened 2021-04-12
Call update/remove Slack message methods without `as_user` param for V2 bot token

zmarushchak-hs:
Slack fails to execute method if `as_user` param is present and V2 bot token is used. Models updated to have the possibility to not include this param for `chat.delete` and `chat.update` methods.


slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/AbstractChatDeleteParams.java
```diff
@@ -7,18 +7,11 @@
...
-  @Override
-  @JsonProperty("channel")
-  public abstract String getChannelId();
-
-  @JsonProperty("ts")
-  public abstract String getMessageToDeleteTs();
+public abstract class AbstractChatDeleteParams extends AbstractChatDeleteParamsBase {
 
   @Default
   @JsonProperty("as_user")
```
omotnyk, line 17: The current solution is 👍 but I wonder what do you think about annotating this property with `@JsonInclude(Include.NON_DEFAULT)`? This annotation says that if the value is the default we won't be serializing the property.

zmarushchak-hs: Good point, didn't know about thit annotation.

PR #225, opened 2021-04-14
Add integration test for ChatDeleteParams model

zmarushchak-hs:
Add integration test for `ChatDeleteParams` model to ensure that `as_user` param is not sent to Slack if it contains default value (`false`).
Slack doesn't allow to include `as_user` param in case if requests are executed by new granular bot.


leonidhladii:
LGTM, all tests are passed

PR #231, opened 2021-06-02
Fix 'No serializer found for class java.lang.Object' on getTeamInfo method execution

zmarushchak-hs:
`ObjectMapper` fails to serialize an instance of class `Object`. Because of this `getTeamInfo` method fails with 'No serializer found for class java.lang.Object'. since it passes `new Object()` as a `params` field to the `postSlackCommand` method. This PR fixes this by passing an empty map instead of the object's instance.


leonidhladii:
LGTM

PR #232, opened 2021-06-03
Add team name to SlackTeam model

zmarushchak-hs:
Added Slack team name to `SlackTeam` model to be able to get team name with [team.info](https://api.slack.com/methods/team.info#response) method.
Also, the `team.info` response doesn't contain the property `slack_team`. The corresponding JSON property is named `team`. This was fixed in the `TeamInfoResponse` model.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/teams/SlackTeamIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/teams/SlackTeamIF.java
@@ -13,7 +13,7 @@ import com.hubspot.immutables.style.HubSpotStyle;
 public interface SlackTeamIF {
   String getId();
   String getDomain();
-  String getName();
+  Optional<String> getName();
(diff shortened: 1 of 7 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/teams/SlackTeamIF.java
```diff
@@ -13,6 +13,7 @@
 public interface SlackTeamIF {
   String getId();
   String getDomain();
+  String getName();
```
omotnyk, line 16: I think this should be `Optional`. This client is by lots of projects so if we add a mandatory field, we must be ready to migrate all of them or make sure nobody does something like `SlackTeam.builder().setId("team_id").setDomain("domain").build()`. Otherwise, we are introducing backward-incompatible change.

omotnyk (changes requested):

omotnyk (approved):

PR #233, opened 2021-06-17
Add Icon field to SlackTeam model

zmarushchak-hs:
I need to get the icon URL for the Slack team. Added Icon field to the SlackTeam model.


leonidhladii (approved):
LGTM

PR #235, opened 2021-06-28
Add method for Slack API's migration.exchange endpoint.

higityan:
Add method for Slack API's migration.exchange endpoint. After migration slack ids has changed, this endpoint returns map of old to new ids.


lmargaryan (commented):
LGTM

@omotnyk just asking you to have another look at this in case I've missed something. Thanks!

omotnyk:
Taking a look now 👀

slack-base/src/main/java/com/hubspot/slack/client/methods/params/migration/MigrationExchangeParamsIF.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+
+@Value.Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface MigrationExchangeParamsIF {
+  List<String> getUsers();
```
omotnyk, line 16: Does it make sense to have this as a `Set`?

omotnyk: The method can be renamed to `getUserIds` to better reveal the intention. You can add `@JsonProperty("users")` then to make sure it's compatible with the API.

slack-base/src/main/java/com/hubspot/slack/client/models/response/migration/MigrationExchangeResponseIF.java
```diff
@@ -0,0 +1,25 @@
...
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface MigrationExchangeResponseIF extends SlackResponse {
+  String getTeamId();
+
+  String getEnterpriseId();
+
+  Map<String, String> getUserIdMap();
+
+  Optional<List<String>> getInvalidUserIds();
```
omotnyk, line 24: Does Slack always return this array of user IDs? I'm curious if we need `Optional` here? How does the response look like if all the IDs are valid? Is it `"invalid_user_ids": []` or there is no `invalid_user_ids` property at all?

slack-base/src/main/java/com/hubspot/slack/client/models/response/migration/MigrationExchangeResponseIF.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface MigrationExchangeResponseIF extends SlackResponse {
+  String getTeamId();
+
+  String getEnterpriseId();
+
+  Map<String, String> getUserIdMap();
```
omotnyk, line 22: What is the format of the response in case all user IDs are invalid? Is it `"user_id_map": {}` or no property is included in the response at all?

omotnyk (commented):
Left a few comments/questions. Also, I'd appreciate adding a few small integration tests like [this one](https://github.com/HubSpot/slack-client/blob/d1e1169ba6e3f86acd08c30a7889057e22f12b3b/slack-base/src/test/java/com/hubspot/slack/client/models/users/SlackUserDeserializationTest.java) to check if request/response models are serialized/deserialized correctly.

slack-base/src/main/java/com/hubspot/slack/client/models/response/migration/MigrationExchangeResponseIF.java
higityan: When there is invalid user ids slack returns this set of ids. But when all ids are valid slack doesn't return this set at all that's why it is Optional.

slack-base/src/main/java/com/hubspot/slack/client/models/response/migration/MigrationExchangeResponseIF.java
higityan: In case of user id map, slack always returns this map, and when all ids are invalid in slack response we get empty map.

slack-base/src/main/java/com/hubspot/slack/client/methods/params/migration/MigrationExchangeParamsIF.java
higityan: Yes makes sense. Let me do that changes.

higityan: Updated.

higityan:
Yes, I'm adding them. Will push them in a while.

higityan:
Updated.

slack-base/src/main/java/com/hubspot/slack/client/models/response/migration/MigrationExchangeResponseIF.java
```diff
@@ -0,0 +1,25 @@
...
+
+  Optional<Set<String>> getInvalidUserIds();
```
omotnyk, line 24: One more thing. I wonder if it's easier to have `Set<String>` instead of `Optional` here. `Set<String> getInvalidUserIds` will be parsed as an empty set if no property is present in JSON at all. So we still can work with the entity but IMHO, it's a bit easier to work with a set that can be empty than with `Optional`. What do you think?

omotnyk (commented):
LGTM. Dropped one small comment regarding the optional usage.

slack-base/src/main/java/com/hubspot/slack/client/models/response/migration/MigrationExchangeResponseIF.java
higityan: Yes that will be better. I just didn't now that json can deserialize without exception. when set property is missing. I'll update it.

higityan: Updated.

slack-base/src/test/java/com/hubspot/slack/client/models/migration/MigrationExchangeParamsDeserializationTest.java
```diff
@@ -0,0 +1,64 @@
...
+  private MigrationExchangeResponse fetchAndDeserializeMigrationExchangeResponse(String jsonFileName) throws IOException {
+    String rawJson = JsonLoader.loadJsonFromFile(jsonFileName);
+    return ObjectMapperUtils.mapper().readValue(rawJson, MigrationExchangeResponse.class);
+  }
+
+  private boolean doesMainParamsPresent(MigrationExchangeResponse migrationExchangeResponse) {
+    return migrationExchangeResponse.isOk() &&
+        !migrationExchangeResponse.getTeamId().isEmpty() &&
+        !migrationExchangeResponse.getEnterpriseId().isEmpty();
+  }
```
omotnyk, line 63: I think it makes sense to change this to
```java
private void assertMainResponseFields(MigrationExchangeResponse migrationExchangeResponse) {
    assertThat(migrationExchangeResponse.isOk()).isTrue();
    assertThat(migrationExchangeResponse.getTeamId()).isEqualTo("T024G0P55");
    assertThat(migrationExchangeResponse.getEnterpriseId()).isEqualTo("E01L6J4CPS8");
  }
```
This way, we verify each field separately and will receive more meaningful error in case of failure. Also, we'll verify that the values from the json fixture are parsed correctly instead of simply checking that they are present.

slack-base/src/test/java/com/hubspot/slack/client/models/migration/MigrationExchangeParamsDeserializationTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Test
+  public void itDeserializesMigrationExchangeResponseWithValidAndinvalidUserIds() throws IOException {
+    MigrationExchangeResponse migrationExchangeResponse =
+        fetchAndDeserializeMigrationExchangeResponse("migration_exchange_response_with_valid_and_invalid_user_ids.json");
+    Map<String, String> userIdMap = migrationExchangeResponse.getUserIdMap();
+    Set<String> invalidUserIds = migrationExchangeResponse.getInvalidUserIds();
+    assertTrue(doesMainParamsPresent(migrationExchangeResponse));
+    assertTrue(userIdMap.containsKey(VALID_USER_OLD_ID) && userIdMap.containsValue(VALID_USER_MIGRATED_ID));
+    assertTrue(invalidUserIds.contains(INVALID_USER_ID));
```
omotnyk, line 29: This looks a bit better to me:
```java
    assertMainResponseFields(migrationExchangeResponse);
    assertThat(userIdMap).containsKey(VALID_USER_OLD_ID);
    assertThat(userIdMap).containsValue(VALID_USER_MIGRATED_ID);
    assertThat(invalidUserIds).containsExactly(INVALID_USER_ID);
```
What do you think?

slack-base/src/test/java/com/hubspot/slack/client/models/migration/MigrationExchangeParamsDeserializationTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Test
+  public void itDeserializesMigrationExchangeResponseWithOnlyValidUserIds() throws IOException {
+    MigrationExchangeResponse migrationExchangeResponse =
+        fetchAndDeserializeMigrationExchangeResponse("migration_exchange_response_with_only_valid_user_ids.json");
+    Map<String, String> userIdMap = migrationExchangeResponse.getUserIdMap();
+    Set<String> invalidUserIds = migrationExchangeResponse.getInvalidUserIds();
+    assertTrue(doesMainParamsPresent(migrationExchangeResponse));
+    assertTrue(userIdMap.containsKey(VALID_USER_OLD_ID) && userIdMap.containsValue(VALID_USER_MIGRATED_ID));
+    assertTrue(invalidUserIds.isEmpty());
```
omotnyk, line 40: This looks a bit better to me:
```java
    assertMainResponseFields(migrationExchangeResponse);
    assertThat(userIdMap).containsKey(VALID_USER_OLD_ID);
    assertThat(userIdMap).containsValue(VALID_USER_MIGRATED_ID);
    assertThat(invalidUserIds).isEmpty();
```
What do you think?

slack-base/src/test/java/com/hubspot/slack/client/models/migration/MigrationExchangeParamsDeserializationTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Test
+  public void itDeserializesMigrationExchangeResponseWithOnlyInvalidUserIds() throws IOException {
+    MigrationExchangeResponse migrationExchangeResponse =
+        fetchAndDeserializeMigrationExchangeResponse("migration_exchange_response_with_only_invalid_user_ids.json");
+    Map<String, String> userIdMap = migrationExchangeResponse.getUserIdMap();
+    Set<String> invalidUserIds = migrationExchangeResponse.getInvalidUserIds();
+    assertTrue(doesMainParamsPresent(migrationExchangeResponse));
+    assertTrue(userIdMap.isEmpty());
+    assertTrue(invalidUserIds.contains(INVALID_USER_ID));
```
omotnyk, line 51: This looks a bit better to me:
```java
    assertMainResponseFields(migrationExchangeResponse);
    assertThat(userIdMap).isEmpty();
    assertThat(invalidUserIds).containsExactly(INVALID_USER_ID);
```
What do you think?

slack-base/src/test/java/com/hubspot/slack/client/models/migration/MigrationExchangeParamsDeserializationTest.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import org.junit.Test;
+
+import com.hubspot.slack.client.jackson.ObjectMapperUtils;
+import com.hubspot.slack.client.models.JsonLoader;
+import com.hubspot.slack.client.models.response.migration.MigrationExchangeResponse;
+
+public class MigrationExchangeParamsDeserializationTest {
```
omotnyk, line 15: The first three comments in this file depend on the last one.

omotnyk: Also, you'll need to use `import static org.assertj.core.api.Assertions.assertThat;` in order to change the assertions as I suggested.

omotnyk (approved):
Now that we've removed `Optional`, I think it's okay to rewrite the tests to make them easier to read/understand. I suggest using `AssertJ` for assertions and the lib is already connected to this project.

Overall, LGTM. Please, update the tests if you think my comments make sense and feel free to merge without further approval.

Thank you!

slack-base/src/test/java/com/hubspot/slack/client/models/migration/MigrationExchangeParamsDeserializationTest.java
higityan: Yes with assertThat it will return more precise errors.

slack-base/src/test/java/com/hubspot/slack/client/models/migration/MigrationExchangeParamsDeserializationTest.java
higityan: Thank you!

lmargaryan:
🚢

higityan:
I updated the test to use there assertj according to comments. Will merge it. 
Thank you!

PR #236, opened 2021-06-30
Changing migration exchange params to use as userIds String  separated with commas in place of set.

higityan:
Slack needed string separated with comma in place of set for migration exchange endpoint. And deleting serialization test because of default method in Param immutable object (Reading json can't misses users property).


lmargaryan:
🚢

PR #238, opened 2021-07-23
Add support of conversations.inviteShared Slack method

zmarushchak-hs:
This PR adds support for [conversations.inviteShared](https://api.slack.com/methods/conversations.inviteShared) Slack method used for invitation of external users to the shared channel. In fact, this method can be used for Slack shared channel creation by invitation users from external workspace to common Slack channel.


leonidhladii:
Looks good to me

PR #239, opened 2021-08-09
Add reactions to conversation history

aybbr:
Added reactions to the `LiteMessage` model to be able to retrieve the list or reactions attached to a slack message.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
@@ -3,6 +3,8 @@ package com.hubspot.slack.client.models;
 import java.util.List;
 
 import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.hubspot.immutables.style.HubSpotStyle;
@@ -11,9 +13,10 @@ import com.hubspot.immutables.style.HubSpotStyle;
 @HubSpotStyle
 @JsonNaming(SnakeCaseStrategy.class)
 public interface ReactionIF {
+  @JsonProperty("name")
   String getName();
-
+  @JsonProperty("count")
   Integer getCount();
-
+  @JsonProperty("users")
   List<String> getUser();
 }
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
@@ -1,0 +1,19 @@
+package com.hubspot.slack.client.models;
+
+import java.util.List;
+
+import org.immutables.value.Value.Immutable;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ReactionIF {
+  String getName();
+
+  Integer getCount();
+
+  List<String> getUser();
+}
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
@@ -4,7 +4,6 @@ import java.util.List;
 
 import org.immutables.value.Value.Immutable;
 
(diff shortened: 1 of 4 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
```diff
@@ -0,0 +1,22 @@
...
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ReactionIF {
+  @JsonProperty("name")
+  String getName();
+  @JsonProperty("count")
+  Integer getCount();
+  @JsonProperty("users")
+  List<String> getUser();
```
omotnyk, line 21: Should we rename this to `getUsers`? I imagine the list is returned when multiple users reacted with the same emoji.

omotnyk: Also, we won't need `JsonProperty` annotation in this case.

slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
```diff
@@ -0,0 +1,22 @@
...
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ReactionIF {
+  @JsonProperty("name")
+  String getName();
+  @JsonProperty("count")
+  Integer getCount();
+  @JsonProperty("users")
```
omotnyk, line 20: We tend to use `@JsonProperty` only if the name of the method is different from JSON key. 
In this case, we don't need to name each property explicitly since for example `count` is automatically deserialized and stored in `getCount`. That's the magic of the immutables.

slack-base/src/main/java/com/hubspot/slack/client/models/ReactionIF.java
```diff
@@ -0,0 +1,22 @@
...
+import com.hubspot.immutables.style.HubSpotStyle;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ReactionIF {
+  @JsonProperty("name")
+  String getName();
+  @JsonProperty("count")
+  Integer getCount();
```
omotnyk, line 19: `Integer` -> `int`

slack-base/src/test/resources/conversations.replies.with_replies_and_new_fields.json
```diff
@@ -35,7 +35,19 @@
...
+          "name": "astonished",
+          "count": 3,
+          "users": [ "U061F7AUR", "U062F7AUR", "U063F7AUR"]
+        },
+        {
+          "name": "facepalm",
+          "count": 34,
+          "users": [ "U061F7AUR", "U062F7AUR", "U063F7AUR", "U064F7AUR", "U065F7AUR" ]
+        }
+      ]
```
omotnyk, line 50: Thanks for adding these. 
Can we also please make sure that these reactions are parsed in the [corresponding integration test](https://github.com/HubSpot/slack-client/blob/d73e725448f8a0f59c5294639ca421ca87f05e32/slack-base/src/test/java/com/hubspot/slack/client/models/response/conversations/ConversationsRepliesResponseTest.java#L27-L36)?

omotnyk (commented):
Good job👍 
I left a few small comments.

aybbr:
Thanks for the quick feedback @omotnyk. I addressed the previous comments so feel free to take a second look.

slack-base/src/test/java/com/hubspot/slack/client/models/response/conversations/ConversationsRepliesResponseTest.java
```diff
@@ -33,6 +33,10 @@ public void itDeserializesWithRepliesAndNewFields() throws IOException {
     assertThat(response.getMessages().get(0).getReplyUserIds().get()).isEqualTo(Arrays.asList("U061F7AUR", "U061F7AUR", "U061F7AUR"));
     assertThat(response.getMessages().get(0).getReplyUsersCount().get()).isEqualTo(3);
     assertThat(response.getMessages().get(0).getLatestReplyTimestamp().get()).isEqualTo("1483125339.020269");
+    assertThat(response.getMessages().get(1).getReactions().size()).isEqualTo(2);
+    assertThat(response.getMessages().get(1).getReactions().get(0).getName()).isEqualTo("astonished");
+    assertThat(response.getMessages().get(1).getReactions().get(0).getCount()).isEqualTo(3);
+    assertThat(response.getMessages().get(3).getReactions().get(1).getUsers()).isEqualTo(Arrays.asList("U061F7AUR", "U062F7AUR", "U063F7AUR"));
```
omotnyk, line 39: Let's either assert that `facepalm` reaction is deserialized correctly here as well or remove it from the JSON fixture.

omotnyk (approved):
Thanks for all your work. I have one small comment left but everything LGTM 🚢

PR #240, opened 2021-08-09
Add schedule message endpoints

higityan:
This pr adds schedule message endpoints and unit test for them.
 1. chat.scheduleMessage
 2. chat.deleteScheduledMessage
 3. chat.scheduledMessages.list
adds long in ReflectionBasedFieldPresenceTest's numeric parameter type checking.


slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/ChatDeleteScheduledMessageParamsIF.java
```diff
@@ -0,0 +1,25 @@
...
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface ChatDeleteScheduledMessageParamsIF extends HasChannel {
+  @Override
+  @JsonProperty("channel")
+  String getChannelId();
+
+  String getScheduledMessageId();
+
+  @JsonProperty("as_user")
```
zmarushchak-hs, line 21: Do we need this? Looks like [the argument](https://api.slack.com/methods/chat.deleteScheduledMessage#arg_as_user) looks the same.

slack-base/src/main/java/com/hubspot/slack/client/models/response/chat/ChatDeleteScheduledMessageResponseIF.java
```diff
@@ -0,0 +1,14 @@
...
+import org.immutables.value.Value.Immutable;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import com.hubspot.slack.client.models.response.SlackResponse;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
```
zmarushchak-hs, line 12: Do we need this?

zmarushchak-hs (approved):
Just a few nits. Please feel free to merge after they are addressed.

slack-base/src/main/java/com/hubspot/slack/client/methods/params/chat/ChatDeleteScheduledMessageParamsIF.java
higityan: Yes, it's fully the same, I'll update and fix it. Thank you.

slack-base/src/main/java/com/hubspot/slack/client/models/response/chat/ChatDeleteScheduledMessageResponseIF.java
higityan: Right there is no need for that, because there is no response metadata in this endpoint's response. I'll update and fix it.

PR #251, opened 2021-10-28
Add method for conversation kick api.

higityan:
Adding method for conversation kick api and unit tests for serializing/deserializing that jsons.


slack-base/src/test/resources/conversation_kick_response.json
```diff
@@ -0,0 +1,6 @@
+{
+  "ok": true,
+  "response_metadata": {
+    "next_cursor": ""
```
omotnyk, line 4: QQ, why do we have this param in place? Is it to test that the model still works when the response contains a new unexpected field?

omotnyk (approved):
Thanks for adding the corresponding UTs 👍 
LGMT 🚀
I left one small question.

slack-base/src/test/resources/conversation_kick_response.json
higityan: Actually this field is in SlackResponse form which every response classes extend, but in ConversationKickResponse we don't need that.

PR #253, opened 2021-11-09
Add subtype for conversations-select input

zmarushchak-hs:
Added [fields](https://api.slack.com/surfaces/modals/using#modal_response_url) for `ConversationsSelectMenu` model to receive channel id in the View Submission. Also added subtype of `ViewInput` type for the conversations-select input object unmarshalling.

Context: when the user selects the channel from [Conversations list](https://api.slack.com/reference/block-kit/block-elements#conversation_select) on the model, Slack sends are `conversations-select` object in the view submission `state` field (please see example below). In order to get the select channel, we need to have a corresponding model.
```
{
  "type": "view_submission",
    .
    .
    .
    "callback_id": "callbackId",
    "state": {
      "values": {
       .
       .
       .
        "cdA=j": {
          "conversation-select": {
            "type": "conversations_select",
            "selected_conversation": "CHANNELID"
          }
        }
      }
    },
    "hash": "hash",
    "title": {
```

### DIFF
--- a/slack-base/pom.xml
+++ b/slack-base/pom.xml
@@ -11,10 +11,10 @@
   <artifactId>slack-base</artifactId>
 
   <dependencies>
-    <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
-    </dependency>
+<!--    <dependency>-->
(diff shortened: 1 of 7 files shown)


leonidhladii (approved):
LGTM 🚢

PR #254, opened 2021-11-16
Add view submission model with callback id

zmarushchak-hs:
Added model for [View Submission ](https://api.slack.com/reference/interaction-payloads/views#view_submission) payload. The existing [model](https://github.com/HubSpot/slack-client/blob/master/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/ViewSubmissionIF.java) assumes that callbak id will never come, so `null` is always set for the callback id field. Not sure the reason for such implementation, but a callback id can be sent in view submission. I didn't want to change the existing model to return Optional<String> for the callback id in order to not break exisying code.


omotnyk (commented):
Can't we use [this](https://immutables.github.io/immutable.html#nullable-attributes) instead and delete the default implementation?

omotnyk (approved):

PR #256, opened 2021-12-06
Normalize Slack form element presentational data

omotnyk:
I implemented normalization of the data that serves informational purposes - label, hint, a placeholder. We don't want to normalize stuff like `name` or `value` because their values can be used as unique identifiers and this can break the consumer in a very unexpected way.

Example of previous behavior: Let's assume that the element label is limited to 24 characters. If we set `aaaaaaaaaaaaaaaaaaaaaaaaaa`(26 chars long) as a label, the model is going to fail on the build.
Current behavior: We are going to show `aaaaaaaaaaaaaaaaaaaaaa...` instead to fit into the limit.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/AbstractSlackFormSelectElement.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/AbstractSlackFormSelectElement.java
@@ -16,13 +16,13 @@ import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.hubspot.immutables.style.HubSpotStyle;
 import com.hubspot.slack.client.models.actions.SlackDataSource;
 import com.hubspot.slack.client.models.dialog.form.SlackFormElementTypes;
-import com.hubspot.slack.client.models.dialog.form.elements.helper.SlackDialogElementNormalizer;
+import com.hubspot.slack.client.models.dialog.form.elements.helpers.SlackDialogElementNormalizer;
 
 @Immutable
 @HubSpotStyle
 @JsonNaming(SnakeCaseStrategy.class)
 @JsonInclude(Include.NON_EMPTY)
-public abstract class AbstractSlackFormSelectElement extends SlackDialogFormElement {
+public abstract class AbstractSlackFormSelectElement extends SlackDialogFormElement implements HasOptions {
   @Default
   @Override
   public SlackFormElementTypes getType() {
@@ -34,7 +34,6 @@ public abstract class AbstractSlackFormSelectElement extends SlackDialogFormElem
     return SlackDataSource.STATIC;
   }
 
-  public abstract List<SlackFormOption> getOptions();
   public abstract List<SlackFormOptionGroup> getOptionGroups();
   public abstract Optional<String> getValue();
   public abstract List<SlackFormOption> getSelectedOptions();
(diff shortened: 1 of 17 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/helper/SlackDialogElementNormalizer.java
```diff
@@ -0,0 +1,152 @@
...
+          .withHint(normalizeHint(element));
+    }
+    return element;
+  }
+
+  public static AbstractSlackFormSelectElement normalize(AbstractSlackFormSelectElement element) {
+    if (shouldNormalizePlaceholder(element, SlackDialogFormElementLengthLimits.MAX_PLACEHOLDER_LENGTH)
+        || shouldNormalizeLabel(element, SlackDialogFormElementLengthLimits.MAX_LABEL_LENGTH)
+        || shouldNormalize(element.getOptionGroups(), SlackDialogFormElementLengthLimits.MAX_OPTION_GROUPS_NUMBER)
+        || shouldNormalize(element.getOptions(), SlackDialogFormElementLengthLimits.MAX_OPTIONS_NUMBER)) {
```
zmarushchak-hs, line 54: The same is duplicated on line 73, so we can extract it to a separate method. But I don't insist on this chanhe.

slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/helper/SlackDialogElementNormalizer.java
```diff
@@ -0,0 +1,152 @@
+package com.hubspot.slack.client.models.dialog.form.elements.helper;
```
zmarushchak-hs, line 1: Should this package be named `helpers`?

slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/AbstractSlackFormSelectElement.java
```diff
@@ -40,56 +41,61 @@ public SlackDataSource getDataSource() {
...
-    int numOptions = getOptions().size();
-    int numOptionGroups = getOptionGroups().size();
+    List<SlackFormOption> normalizedOptions = normalized.getOptions();
+    int numOptions = normalizedOptions.size();
+    List<SlackFormOptionGroup> normalizedOptionGroups = normalized.getOptionGroups();
+    int numOptionGroups = normalizedOptionGroups.size();
 
-    if (numOptions > 100) {
+    if (numOptions > SlackDialogFormElementLengthLimits.MAX_OPTIONS_NUMBER.getLimit()) {
       errors.add("Cannot have more than 100 options");
```
zmarushchak-hs, line 55: Can we use limit values from enum here and in all other logging statements? I think we can update it in the enum and forget to update it here.

zmarushchak-hs (commented):
Left few comments

slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/helper/SlackDialogElementNormalizer.java
```diff
@@ -0,0 +1,152 @@
...
+      return SlackFormOptionGroup.builder()
+          .from(element)
+          .setLabel(normalizeLabel(element))
+          .setOptions(normalizeOptions(element))
+          .build();
+    }
+    return element;
+  }
+
+  private static boolean shouldNormalize(List listOfFormElements, SlackDialogFormElementLengthLimits maxListSize) {
```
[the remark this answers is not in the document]
omotnyk: We can, I simply wanted to keep this super generic so it can be reused in the future.

slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/AbstractSlackFormSelectElement.java
omotnyk: Good call 👍

omotnyk:
@zmarushchak-hs Your comments were addressed. Can you please take another look?

zmarushchak-hs (approved):
LGTM!

PR #258, opened 2021-12-07
Normalize Slack form element presentational data | Attempt 2

omotnyk:
Same as https://github.com/HubSpot/slack-client/pull/256 but fixed stackoverflow error during normalization caused by cyclic constructor calls in `with` methods.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/helpers/SlackDialogElementNormalizer.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/helpers/SlackDialogElementNormalizer.java
@@ -28,10 +28,12 @@ public class SlackDialogElementNormalizer {
     if (shouldNormalizePlaceholder(element)
         || shouldNormalizeLabel(element, SlackDialogFormElementLengthLimits.MAX_LABEL_LENGTH)
         || shouldNormalize(element.getHint(), SlackDialogFormElementLengthLimits.MAX_HINT_LENGTH)) {
-      return SlackFormTextElement.copyOf(element)
-          .withPlaceholder(normalizePlaceholder(element))
-          .withLabel(normalizeLabel(element))
-          .withHint(normalizeHint(element));
+      return SlackFormTextElement.builder()
+          .from(element)
+          .setPlaceholder(normalizePlaceholder(element))
+          .setLabel(normalizeLabel(element))
+          .setHint(normalizeHint(element))
+          .build();
     }
     return element;
   }
@@ -40,10 +42,12 @@ public class SlackDialogElementNormalizer {
     if (shouldNormalizePlaceholder(element)
         || shouldNormalizeLabel(element, SlackDialogFormElementLengthLimits.MAX_LABEL_LENGTH)
         || shouldNormalize(element.getHint(), SlackDialogFormElementLengthLimits.MAX_HINT_LENGTH)) {
-      return SlackFormTextareaElement.copyOf(element)
-          .withPlaceholder(normalizePlaceholder(element))
-          .withLabel(normalizeLabel(element))
-          .withHint(normalizeHint(element));
+      return SlackFormTextareaElement.builder()
+          .from(element)
+          .setPlaceholder(normalizePlaceholder(element))
+          .setLabel(normalizeLabel(element))
+          .setHint(normalizeHint(element))
+          .build();
     }
     return element;
   }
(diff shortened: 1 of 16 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/helpers/SlackDialogElementNormalizer.java
```diff
@@ -0,0 +1,162 @@
...
+  public static AbstractSlackFormTextElement normalize(AbstractSlackFormTextElement element) {
+    if (shouldNormalizePlaceholder(element)
+        || shouldNormalizeLabel(element, SlackDialogFormElementLengthLimits.MAX_LABEL_LENGTH)
+        || shouldNormalize(element.getHint(), SlackDialogFormElementLengthLimits.MAX_HINT_LENGTH)) {
+      return SlackFormTextElement.builder()
+          .from(element)
+          .setPlaceholder(normalizePlaceholder(element))
+          .setLabel(normalizeLabel(element))
+          .setHint(normalizeHint(element))
+          .build();
```
omotnyk, line 36: Replaced `with` calls with builder here.

slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/helpers/SlackDialogElementNormalizer.java
```diff
@@ -0,0 +1,162 @@
...
+  public static AbstractSlackFormTextareaElement normalize(AbstractSlackFormTextareaElement element) {
+    if (shouldNormalizePlaceholder(element)
+        || shouldNormalizeLabel(element, SlackDialogFormElementLengthLimits.MAX_LABEL_LENGTH)
+        || shouldNormalize(element.getHint(), SlackDialogFormElementLengthLimits.MAX_HINT_LENGTH)) {
+      return SlackFormTextareaElement.builder()
+          .from(element)
+          .setPlaceholder(normalizePlaceholder(element))
+          .setLabel(normalizeLabel(element))
+          .setHint(normalizeHint(element))
+          .build();
```
omotnyk, line 50: Replaced `with` calls with builder here.

slack-base/src/main/java/com/hubspot/slack/client/models/dialog/form/elements/helpers/SlackDialogElementNormalizer.java
```diff
@@ -0,0 +1,162 @@
...
+        || shouldNormalizeLabel(element, SlackDialogFormElementLengthLimits.MAX_LABEL_LENGTH)
+        || shouldNormalize(element.getOptionGroups(), SlackDialogFormElementLengthLimits.MAX_OPTION_GROUPS_NUMBER)
+        || shouldNormalizeOptions(element)) {
+      return SlackFormSelectElement.builder()
+          .from(element)
+          .setPlaceholder(normalizePlaceholder(element))
+          .setLabel(normalizeLabel(element))
+          .setOptionGroups(normalizeOptionGroups(element))
+          .setOptions(normalizeOptions(element))
+          .build();
```
omotnyk, line 66: Replaced `with` calls with builder here.

omotnyk, line 66: I'll describe the stack overflow problem with an example. Let's say we have the old code here:
```java
    if (shouldNormalizePlaceholder(element)
        || shouldNormalizeLabel(element, SlackDialogFormElementLengthLimits.MAX_LABEL_LENGTH)
        || shouldNormalize(element.getHint(), SlackDialogFormElementLengthLimits.MAX_HINT_LENGTH)) {
      return SlackFormTextElement.copyOf(element)
          .withPlaceholder(normalizePlaceholder(element))
          .withLabel(normalizeLabel(element))
          .withHint(normalizeHint(element));
    }
    return element;
```

What we do here is define whether we should normalize any data first and then normalize it. The problem is actually obvious but still tricky. Let's say we have a very long `label` that should be normalized. In this case, when `SlackFormSelectElement` is created we do next:
1. Check if we should normalize any field.
2. Determine that we should normalize some fields(`label`).
3. Start normalization process:
    1. Normalize placeholer and then call `withPlaceholder`.
    2. Normalize label and then call `withLabel`.
    3. Normalize the rest.
    4. Build the final normalized object.

The flow I described above is what I think is happening. What was actually happening is we were entering an endless loop upon calling `withPlaceholder` because unlike builder, this method creates the actual `SlackFormSelectElement` so it calls method annotated with `@Value.Check`, which then initiates the normalization process once again. SInce we didn't normalize the label yet, we determine that `label` still has to be normalized and run `SlackFormTextElement.copyOf(element).withPlaceholder(normalizePlaceholder(element)).withLabel(normalizeLabel(element))` once again, which results in doing this again, and then again, and again.

This issue wasn't caught at first because it depends on the order of the checks and normalization as well as on the data to be normalized. If we place `withLabel` before `withPlaceholder` and try to build the object with a long label, everything works fine because on the call to `withLabel` method we have a fully normalized object that passes all the checks.

zmarushchak-hs (approved):
LGTM!

PR #260, opened 2021-12-09
Add static select view input model

zmarushchak-hs:
Added model to be able to deserialize input provided by the static select element upon view submission.


leonidhladii:
LGTM 🚢

PR #261, opened 2021-12-09
Add dispatch_action field to Input Block

zmarushchak-hs:
`dispatch_action` field should be set to true for the [Input Block](https://api.slack.com/reference/block-kit/blocks#input) in order to be notified about user interaction with this block. E.g.: Slack will notify the app when the user selects an option from the select menu.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/InputIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/InputIF.java
@@ -34,4 +34,6 @@ public interface InputIF extends Block {
 
   @JsonProperty("optional")
   Optional<Boolean> isOptional();
+
+  Optional<Boolean> getDispatchAction();
 }


slack-base/src/main/java/com/hubspot/slack/client/models/blocks/InputIF.java
```diff
@@ -34,4 +34,6 @@ default String getType() {
 
   @JsonProperty("optional")
   Optional<Boolean> isOptional();
+
```
leonidhladii, line 37: Do we need @JsonProperty("dispatch_action") ?

zmarushchak-hs: No, having `@JsonNaming(SnakeCaseStrategy.class)` annotation is enough.

zmarushchak-hs: Checked if it works and it works!

leonidhladii (approved):
🚀

leonidhladii (approved):
🚀

PR #267, opened 2022-02-17
Add getStringValue method to BlockKit inputs

zmarushchak-hs:
Currently, each input type has its own type for the included value it is needed to handle each input separately. It is not convenient when handling a collection of inputs (consider situation when different BlockKit inputs are submitted for the same view). This PR adds a getter to the ViewInput interface to make values extraction more universal (return a String value). 
I decided that converting to String DatePicker and multiple checkboxes inputs makes no sense and I think it is OK to handle them as a special case.


leodevnull (approved):
Good job! Approve 🚀

PR #268, opened 2022-02-21
Normalize option and option group lists in `BlocksLoadOptionsResponse`

omotnyk:
Limit the number of dynamically fetched options in a single place to avoid runtime failures from Slack.


leonidhladii (approved):
🚀

PR #277, opened 2022-04-05
Add SlackOptionsResponse model for Dialog Options handling

zmarushchak-hs:
SlackOptionsResponse model is needed to load Options to [Selects](https://api.slack.com/dialogs#select_elements) on [Slack Dialogs](https://api.slack.com/dialogs).

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/BlocksLoadOptionsResponseIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/BlocksLoadOptionsResponseIF.java
@@ -2,7 +2,6 @@ package com.hubspot.slack.client.models.interaction;
 
 import java.util.List;
 
-import org.immutables.value.Value.Derived;
 import org.immutables.value.Value.Check;
 import org.immutables.value.Value.Immutable;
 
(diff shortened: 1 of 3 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/interaction/BlocksLoadOptionsResponseIF.java
```diff
@@ -19,6 +20,13 @@
 @HubSpotStyle
 @JsonNaming(SnakeCaseStrategy.class)
 public interface BlocksLoadOptionsResponseIF extends BaseSlackOptionsResponse {
+  String TYPE = "blocks_load_options_response";
+
+  @Derived
+  default String getType() {
+    return TYPE;
+  }
```
omotnyk, line 28: Why do we need this?

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/SlackOptionsResponseIF.java
```diff
@@ -0,0 +1,97 @@
...
+public interface SlackOptionsResponseIF extends BaseSlackOptionsResponse {
+  String TYPE = "slack_options_response";
+
+  // NOTE: slack annoyingly has a different model for dialog options and attachment options
+  // https://api.slack.com/dialogs | https://api.slack.com/docs/message-menus
+
+  @Derived
+  default String getType() {
+    return TYPE;
+  }
```
omotnyk, line 36: Why do we need this?

omotnyk (approved):
Two questions. The rest looks good to me.

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/BlocksLoadOptionsResponseIF.java
zmarushchak-hs: Looks like we didn't. I thought it is used by Jackson to determine type but looks like I was wrong. Removing this.

PR #278, opened 2022-04-05
Added missing field parsing

IvanivOleg:
Added parsing of missing "state" field


slack-base/src/main/java/com/hubspot/slack/client/models/interaction/BlockActionsIF.java
```diff
@@ -26,6 +26,8 @@
   @JsonProperty("actions")
   List<BlockElementAction> getElementActions();
 
+  ViewStateValuesPayload getState();
```
omotnyk, line 29: Ideally, we'd like to rename this class to something more generic as it doesn't represent solely "View" state anymore. The only drawback is that this is going to be a breaking change. Thoughts?

slack-base/src/test/resources/block_actions.json
```diff
@@ -75,5 +75,32 @@
...
+                "text":" 2",
+                "emoji":true
+              },
+              "value":"11"
+            }
+          ]
+        }
+      }
+    }
+  }
```
omotnyk, line 105: 👍

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/BlockActionsIF.java
IvanivOleg: Fixed

omotnyk (approved):
🚀

PR #281, opened 2022-04-06
Make `state` in `block_actions` payload optional

omotnyk:
As per [Slack documentation](https://api.slack.com/reference/interaction-payloads/block-actions#fields) it sends either `state.values` or `view.state.values` so both of these fields should be `Optional` so we don't fail do parse the payload if one of them is missing.

`view.state.values` is sent on interactions with blocks inside of modal. 
`state.values` is sent on the interactions with blocks inside of the message.


IvanivOleg (approved):

IvanivOleg:
@omotnyk good catch 🚀

PR #283, opened 2022-04-26
Support `Timepicker` for BlockKit

leonidhladii:
Added support of `TimePicker` for BlockKit. Also added missing `getStringValue` for DatePicker and TimePicker to getting dates value in text format

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/views/ViewDatePickerIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/views/ViewDatePickerIF.java
@@ -1,13 +1,11 @@
 package com.hubspot.slack.client.models.interaction.views;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.hubspot.immutables.style.HubSpotStyle;
 import org.immutables.value.Value;
 
 import java.time.LocalDate;
-import java.time.format.DateTimeFormatter;
 import java.util.Optional;
 
 @Value.Immutable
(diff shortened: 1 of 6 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/interaction/views/ViewDatePickerIF.java
```diff
@@ -1,17 +1,27 @@
...
   Optional<LocalDate> getSelectedDate();
+
+  @JsonIgnore
+  default Optional<String> getStringValue() {
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MM/dd/yy");
+    if (getSelectedDate().isPresent()) {
+      return Optional.of(getSelectedDate().get().format(formatter));
+    }
+    return Optional.empty();
+  }
```
omotnyk, line 26: Two notes here:
1. Why do we need this method?
2. I feel a bit uncertain about the date format. This is US date format and I'm not sure this string is going to be expected in all the cases by all the users of the method. I think in the simplest case we should default to the international format.
3. The date format is completely different in [DatePickerIF](https://github.com/HubSpot/slack-client/blob/dd42993a268b5478caad96942c83169397926b66/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/DatePickerIF.java#L34), which also adds some confusion around this.

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/views/ViewTimePickerIF.java
```diff
@@ -0,0 +1,27 @@
...
+  Optional<LocalTime> getSelectedTime();
+
+  @JsonIgnore
+  default Optional<String> getStringValue() {
+    DateTimeFormatter formatter = DateTimeFormatter.ofPattern("H:mm");
+    if (getSelectedTime().isPresent()) {
+      return Optional.of(getSelectedTime().get().format(formatter));
+    }
+    return Optional.empty();
+  }
```
omotnyk, line 26: Why do we need this method?

slack-base/src/main/java/com/hubspot/slack/client/models/interaction/views/ViewDatePickerIF.java
leonidhladii: 1. we definitely need this method, because we filtered properties without values, you can see this in `DefaultViewInputValueExtractor` and related classes.
2-3. this again related to our date formar parsing, I choose US format as a default string value.

leonidhladii: In general, after our conversation, I agree that to be more flexible we don’t need this method, because each user of this code can have their own date/time format, so I removed `getStringValue` methods from Date and Time pickers.

leonidhladii:
@omotnyk thanks for review, can you take a look again?

PR #293, opened 2022-06-23
Sh/add request verification

dezuz:
### Description
Slack supports request verification mechanisms that are already used in a few projects so it's better to move the logic to this project and then use it wherever is needed. [Slack documentation](https://api.slack.com/authentication/verifying-requests-from-slack)

### Solution
To provide this logic I have done:
- added `slack-request-verifier` module
- added `SlackRequestVerifierFilter` class
- added unit tests to this class

### DIFF
--- a/slack-request-verifier/src/main/java/com/hubspot/slack/client/request/verifier/SlackRequestVerifierFilter.java
+++ b/slack-request-verifier/src/main/java/com/hubspot/slack/client/request/verifier/SlackRequestVerifierFilter.java
@@ -1,0 +1,151 @@
+package com.hubspot.slack.client.request.verifier;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.List;
+
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+
+import org.apache.commons.codec.binary.Hex;
+
+import com.google.common.base.Charsets;
+import com.google.common.base.Preconditions;
+import com.google.common.hash.Hashing;
+import com.google.common.io.CharStreams;
+import com.google.inject.Inject;
+import com.google.inject.assistedinject.Assisted;
+
+/**
+ * Check for detailed information https://api.slack.com/authentication/verifying-requests-from-slack
+ */
+public class SlackRequestVerifierFilter implements ContainerRequestFilter {
+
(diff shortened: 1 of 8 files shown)


dezuz:
I found two possible ways to use this class in other projects:
- add a `bundle` and `feature` and register this filter for resources that are under the annotation
- extend this class, mark extended class with `@Provider` annotation and custom annotation that also would be used on the resource method, and bind it

slack-request-verifier/src/test/java/com.hubspot.slack.client.request.verifier/SlackRequestVerifierFilterTest.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.stream.Stream;
+
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.MultivaluedMap;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class SlackRequestVerifierFilterTest {
```
omotnyk, line 20: 👍

slack-request-verifier/src/main/java/com/hubspot/slack/client/request/verifier/SlackRequestVerifierFilter.java
```diff
@@ -0,0 +1,151 @@
...
+  private static final String ENCODED_SIGNATURE_TEMPLATE = "%s=%s";
+  private static final String VERSION_NUMBER = "v0";
+
+  private final long requestExpirationTime;
+  private final String signingSecret;
+
+  public interface Factory {
+    SlackRequestVerifierFilter create(
+        @Assisted("requestExpirationTime") long requestExpirationTime,
+        @Assisted("signingSecret") String signingSecret
```
omotnyk, line 38: IIRC, we don't have to specify `@Assisted` annotations on the factory params.

omotnyk (approved):
Thanks for working on this @dezuz. I left one small comment but overall it LGTM 👍 🚀

slack-request-verifier/src/main/java/com/hubspot/slack/client/request/verifier/SlackRequestVerifierFilter.java
dezuz: We do so in our internal projects so I thought that it works like that.
I will try to remove them

PR #294, opened 2022-06-28
add SlackRequestVerifier annotation

dezuz:
One of the two ways to implement a slack request verifier needs an annotation over the resource method and class that extends `SlackRequestVerifierFilter`, to prevent duplication of the annotations it's better to add this annotation to this project

### DIFF
--- a/slack-request-verifier/src/main/java/com/hubspot/slack/client/request/verifier/SlackRequestVerifier.java
+++ b/slack-request-verifier/src/main/java/com/hubspot/slack/client/request/verifier/SlackRequestVerifier.java
@@ -1,0 +1,11 @@
+package com.hubspot.slack.client.request.verifier;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import com.google.inject.BindingAnnotation;
+
+@BindingAnnotation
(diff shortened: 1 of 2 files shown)


slack-request-verifier/src/main/java/com/hubspot/slack/client/request/verifier/SlackRequestVerifier.java
```diff
@@ -0,0 +1,11 @@
+package com.hubspot.slack.client.request.verifier;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+import com.google.inject.BindingAnnotation;
+
+@BindingAnnotation
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SlackRequestVerifier {
```
omotnyk, line 10: I feel like `SlackRequestVerifier` is a good name for the filter implementation but the marker's name could point to the action itself(e.g.`VerifySlackRequest`)?

dezuz: `VerifySlackRequest` sounds good to me too. Is it better to rename this annotation?

PR #295, opened 2022-07-25
DO NOT MERGE/REVIEW. Add shared channels filtering.

nliutyi-hs:
In this PR we add filtering logic to filter out shared channels, in case when ConversationsListParams.shouldExcludeSharedChannels() is true.

### DIFF
--- a/slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
+++ b/slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Multimap;
 import com.google.inject.assistedinject.Assisted;
@@ -897,8 +898,7 @@ public class SlackWebClient implements SlackClient {
         );
 
         CompletableFuture<Result<List<Conversation>, SlackError>> pageFuture = resultFuture.thenApply(
-          result -> result.mapOk(ConversationListResponse::getConversations)
-        );
+          result -> result.mapOk(response -> getConversations(response, params)));
 
         CompletableFuture<Optional<String>> nextCursorMaybeFuture = extractNextCursor(
           resultFuture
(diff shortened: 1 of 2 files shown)


slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -897,8 +898,7 @@ protected LazyLoadingPage<Result<List<Conversation>, SlackError>, String> getPag
         );
 
         CompletableFuture<Result<List<Conversation>, SlackError>> pageFuture = resultFuture.thenApply(
-          result -> result.mapOk(ConversationListResponse::getConversations)
-        );
+          result -> result.mapOk(response -> getConversations(response, params)));
```
omotnyk, line 901: 1. We don't really need to pass the whole `params` object to this method. 
2. We can store `params.shouldExcludeSharedChannels().orElse(false)` in a variable once even before we create `AbstractPagedIterable` and pass it to `getConversations`.

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
```diff
@@ -915,6 +915,20 @@ protected LazyLoadingPage<Result<List<Conversation>, SlackError>, String> getPag
     };
   }
 
+  private List<Conversation> getConversations(ConversationListResponse conversationListResponse, ConversationsListParams params) {
+    List<Conversation> conversations = conversationListResponse.getConversations();
+    if (params.shouldExcludeSharedChannels().orElse(false)) {
+      return filterSharedConversations(conversations);
```
omotnyk, line 921: I guess the correct name for this method would be `filterOutShareConversations` because we are filtering all the conversations to remove the shared ones.

omotnyk (commented):
I wonder if we should add any custom filters that do not exist in Slack to this Slack client instead of keeping it as close to Slack API spec as possible? My concern is that it can be confusing for the consumer of this API to see `shouldExcludeSharedChannels` parameter that isn't descirbed in [the official docs](https://api.slack.com/methods/conversations.list).
My secondary concern is that consumers are going to see pages with different sizes depending on the number of shared channels in the initial page.

That said, I'm not sure whether this is a bad idea in general because these is going to allow. What do you think @nliutyi-hs?

slack-java-client/src/main/java/com/hubspot/slack/client/SlackWebClient.java
nliutyi-hs: I've been thinking about this naming too.

nliutyi-hs:
@omotnyk So your main concern is to better leave the slack-client unchanged, thus having the filtering logic in the `slack-integration`.

nliutyi-hs:
Taking into account we anyway need to filter stuff from the cache, maybe it makes more sense.

PR #297, opened 2022-08-15
Added container to block actions

rlmacen:
### Updates

Added Optional `container` field to BlockActions immutable object

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/BlockActionsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/BlockActionsIF.java
@@ -27,7 +27,7 @@ public interface BlockActionsIF extends SlackInteractiveCallback {
 
   Optional<StateValuesPayload> getState();
 
-  Optional<MessageContainer> getContainer();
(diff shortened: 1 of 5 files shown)


omotnyk (commented):
@rlmacen According to [Slack docs](https://api.slack.com/reference/interaction-payloads/block-actions#:~:text=A%20block_actions%20payload%20is%20received,and%20respond%20to%20these%20payloads.) `container` can have different forms so I wonder what's going to happen in case we receive a payload with a container of `view` type like this one? Will the deserialization fail entirely?
image

If it's going to fail, I guess we should account for two of those types to come and deserialize both, or at the very least, if deserialization is too complex, we should safely ignore the container of `view` type. Although, I'd prefer the former option.

rlmacen:
Hey, @omotnyk, thank you for bringing that up! I have replaced MessageContainer with [Container](https://github.com/HubSpot/slack-client/blob/master/slack-base/src/main/java/com/hubspot/slack/client/models/interaction/Container.java) that can be transformed into a message or view container

omotnyk (approved):
👍 🚀 👍 🚀 👍 🚀

PR #304, opened 2022-11-17
Add blocks for messages.

nliutyi-hs:
This PR adds `message blocks`, which are passed by Slack in the `message_action` payload.
This will allow us to work with user messages in a much more sufficient way, removing the need to parse the Slack markdown text itself, but giving the flexibility to work with blocks.

We've used this code version in our QA env and there were no exceptions regarding to it in our runtime.


slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/RichBlockStyle.java
```diff
@@ -0,0 +1,14 @@
+package com.hubspot.slack.client.models.blocks.messages;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public enum RichBlockStyle {
```
motnyk, line 5: This looks like `ListType` to me. `RichBlockStyle` suggests that this can be applied to any rich block but that isn't the case.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/RichMessageBlock.java
```diff
@@ -0,0 +1,9 @@
+package com.hubspot.slack.client.models.blocks.messages;
+
+import com.hubspot.slack.client.models.blocks.Block;
+
+import java.util.List;
+
+public interface RichMessageBlock extends Block {
```
motnyk, line 7: I think that the block with the nested blocks is not a unique feature of message blocks. I'd probably renamed this to `HasNestedBlocks` or `BlockWithNestedBlocks`? Probably the former option is more common.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/RichTextListIF.java
```diff
@@ -0,0 +1,24 @@
...
+    @Value.Derived
+    default String getType() {
+        return TYPE;
+    }
+
+    RichBlockStyle getStyle();
+
+    int getIndent();
+
+    int getBorder();
```
motnyk, line 23: What does it represent? I'm curious about the difference between 1 and 2 for example.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/TextStyleIF.java
```diff
@@ -0,0 +1,23 @@
...
+@Value.Immutable
+@HubSpotStyle
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public interface TextStyleIF {
+    @JsonProperty("bold")
+    Optional<Boolean> isBold();
+    @JsonProperty("italic")
+    Optional<Boolean> isItalic();
+    @JsonProperty("strike")
+    Optional<Boolean> isStrike();
```
motnyk, line 20: Let's call it `isStrikethrough`?

motnyk: I don't see a reason to use the short name here.

slack-base/src/test/java/com/hubspot/slack/client/models/blocks/messages/MessageActionSerializationTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+        assertThat(messageAction).isNotNull();
+
+        LiteMessage message = messageAction.getMessage();
+        List<Block> blocks = message.getBlocks();
+        assertThat(blocks.size()).isEqualTo(BLOCKS_COUNT);
+
+        RichMessageBlock firstBlock = (RichMessageBlock) blocks.get(0);
+        List<Block> elements = firstBlock.getElements();
+        assertThat(elements.size()).isEqualTo(RICH_MESSAGE_BLOCKS_COUNT);
```
motnyk, line 32: We parse unknown blocks into a separate "unknown" type, right? If so, does this mean that counting doesn't really test anything and we should compare the whole payload instead?

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/RichMessageBlock.java
motnyk: Also, QQ, I remember there was some failure because of this additional type when we were not able to correctly parse a new type returned by Slack if it isn't mapped to a model in this client.
Is the same thing going to happen now?

motnyk (changes requested):
@nliutyi-hs 👍  
One small suggestion, I'd really appreciate it if you add an example payload to simplify the PR review.

slack-base/src/test/java/com/hubspot/slack/client/models/blocks/messages/MessageActionSerializationTest.java
nliutyi-hs: Makes sense.

nliutyi-hs:
@motnyk Regarding the example payload, I thought that `message_action.json` is a good example, and there is no need to duplicate it, but I should've mentioned this in the PR's description.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/RichBlockStyle.java
nliutyi-hs: Makes sense.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/RichMessageBlock.java
nliutyi-hs: I think yeah, we can rename it to `BlockWithNestedBlocks`. 
The previous errors were related to  `MessageBlock`, which was more of a marker interface. 
I could not rely 100% on our test run, as we had only our services up, and it's not the complete data set to make a conclusion.
We definitely won't merge this PR today( gonna merge it the morning), so we will be able to catch an error if it happens again.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/TextStyleIF.java
nliutyi-hs: Yes, we can do it.

nliutyi-hs:
@motnyk Gonna simplify the payload and build the message blocks in the code, for decent comparison.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/RichTextListIF.java
nliutyi-hs: Also, it's good you asked about this, I've added the optional offset field, which is present in a few `RichTextList` blocks, but not in all of them.

motnyk:
> Regarding the example payload, I thought that message_action.json is a good example, and there is no need to duplicate it, but I should've mentioned this in the PR's description.

That's a good call, I haven't really gotten to it yet at a point when I realized that I missed an example to simplify the review.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/messages/RichMessageBlock.java
motnyk: @nliutyi-hs Can you try and add a nested block with an random unknown type(e.g. blabla) and run a unit test or something? I guess that's something we can easily catch even before we deploy any code.

motnyk (commented):
@nliutyi-hs Okay, so I think we'll be good to go here once you add a test that covers parsing of all the blocks and address https://github.com/HubSpot/slack-client/pull/304#discussion_r1028377841. Actually, I feel like that comment basically translates into "add another test for parsing unknown nested type".

nliutyi-hs:
@motnyk PR is updated, please take a look once you'll be able to.

motnyk (approved):

PR #312, opened 2023-03-31
Add confirmation dialog `style` field

motnyk:
Add support for styling the confirmation dialog. Color schemes for the buttons in these 3 styles:
image

https://api.slack.com/reference/block-kit/composition-objects#confirm


ykaliukh (approved):
LGTM

PR #313, opened 2023-05-10
Add ContextTeamId to SlackEventWrapper.

mbeziazychnyi:
## Description
This PR adds ContextTeamId to SlackEventWrapper. I need this new field to extrude actual teamID for events that happened  in shared channels, since those events contains team ids for both workspaces. I made it optional because I don't want to break all the builds in the world. 

@zmarushchak-hs @omotnyk @motnyk @nliutyi-hs


ykaliukh (approved):
LGTM!

PR #314, opened 2023-06-20
add blocks to attachments

ykaliukh:
For my fix, I need to add the field "blocks" to the field "attachment" according to [this Slack doc](https://api.slack.com/reference/messaging/attachments#legacy_fields). For some reason, Slack API does not process requests as expected if an attachment has the field "blocks" and other fields. In order to solve this problem, a custom serializer has been added.


slack-base/src/test/resources/attachment_with_blocks.json
```diff
@@ -0,0 +1,18 @@
...
+          "alt_text": "alt text"
+        },
+        {
+          "type": "mrkdwn",
+          "text": "<profileUrl543|name123>"
+        }
+      ]
+    }
+  ]
+}
```
zmarushchak-hs, line 18: New line, please.

slack-base/src/test/resources/attachment_without_blocks.json
```diff
@@ -0,0 +1,27 @@
...
+    }
+  ],
+  "footer": null,
+  "footer_icon": null,
+  "thumb_url": null,
+  "callback_id": null,
+  "actions": [],
+  "mrkdwn_in": [],
+  "ts": null
+}
```
zmarushchak-hs, line 27: New line, please.

slack-base/src/test/resources/attachment_without_blocks.json
```diff
@@ -0,0 +1,27 @@
+{
+  "fallback": null,
```
zmarushchak-hs, line 2: Do you think it makes sense to add tests with with non-null fields?

zmarushchak-hs (commented):
@ykaliukh I left few comments.

zmarushchak-hs (approved):
@ykaliukh LGTM! But please be sure to not merge this until the end of WoR.

PR #320, opened 2023-10-27
Add additional file errors (`file_not_found`, `file_deleted`)

rydelaney-hs:
Expanding on the work of https://github.com/HubSpot/slack-client/pull/318 this adds two new file error deserialisers covering `file_not_found` and `file_deleted`. The case of `file_not_found` was found in production causing some endpoints to fail, however `file_deleted` is assumed to be a valid error based on [this api documentation](https://api.slack.com/methods/files.info#:~:text=request%20is%20denied.-,file_deleted,-The%20requested%20file).

This PR also restructures how the error types are compared in the deserialiser, and the inheritance structure of the file error immutables to make this more flexible to add file errors in the future. Open to feedback on this new structure in case there are any better ways to pull it off, the commit messages themselves explain some more of the justification behind the changes. Additional tests have been added to cover the extra cases and all other tests pass so hopefully no regressions from this structure change.


pmusgrave (approved):
Looks good to me

PR #321, opened 2024-01-27
Add support for common spreadsheet file formats

rmarsch:
Been writing a job that generates a spreadsheet report and sends it to a slack channel. Without a proper type it reports in the slack UI as `Binary`. The file still works so this isn't blocking, but seems worthwhile to cover.

These types are in the [supported listing](https://api.slack.com/types/file#types). Let me know if I'm missing any related implementation details required to support this.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
@@ -14,7 +14,7 @@ public enum SlackFileType {
   PNG("png", SlackPngFile.class),
   JAVASCRIPT("javascript", SlackJavaScriptFile.class),
   XLSX("xlsx", SlackXlsxFile.class),
-  XLS("xlsx", SlackXlsFile.class),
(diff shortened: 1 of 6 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/files/SlackFileType.java
```diff
@@ -13,6 +13,8 @@ public enum SlackFileType {
   JPG("jpg", SlackJpgFile.class),
   PNG("png", SlackPngFile.class),
   JAVASCRIPT("javascript", SlackJavaScriptFile.class),
+  XLSX("xlsx", SlackXlsxFile.class),
+  XLS("xlsx", SlackXlsFile.class),
```
stevie400, line 17: copy-pasta?
```suggestion
  XLS("xls", SlackXlsFile.class),
```

stevie400 (commented):
Looks good overall, with a possible typo. It could be worth adding some quick tests to `SlackFileDeserializerTest`

stevie400 (approved):
LGTM

PR #271, opened 2022-03-15
Add STATIC_SELECT input type to BlockKit input types enum

zmarushchak-hs:
STATIC_SELECT value is missing in ViewInputType enums, causing `static_select` input type to be deserialized to `unknown`. Enum updated in this PR, so ViewStaticSelect type can be built for `static_select`.

Error example:
```{"status":"error","message":"Invalid input JSON on line 1, column 3088: Could not resolve type id 'unknown' as a subtype of `com.hubspot.slack.client.models.interaction.views.ViewInput`: known type ids = [checkboxes, conversations_select, datepicker, external_select, multi_static_select, plain_text_input, radio_buttons, static_select, users_select] (for POJO property 'values')","correlationId":"12be0c36-06a8-457e-a73a-a6f98c463169"}```

PR #322, opened 2024-03-08
Add additional view input types

kanelson-hs:
The following values are missing from the `ViewInputType` enum, causing those input types to be deserialized to "unknown". The enum is updated here so that the `ViewInput` objects can actually be built.

Note that `ViewInput` already includes the types `ViewEmailInput`, `ViewUrlInput`, `ViewNumberInput`, and `ViewDateTimePicker`.

Similar issue to https://github.com/HubSpot/slack-client/pull/271


rydelaney-hs (approved):
Looks good, if needed for future reference

`EMAIL_TEXT_INPUT`: https://api.slack.com/reference/block-kit/block-elements#email  
`URL_TEXT_INPUT`: https://api.slack.com/reference/block-kit/block-elements#url  
`NUMBER_INPUT`: https://api.slack.com/reference/block-kit/block-elements#number  
`DATETIMEPICKER`: https://api.slack.com/reference/block-kit/block-elements#datetimepicker

PR #326, opened 2024-04-04
Add files to slack event model

ykaliukh:
Currently, `SlackEventMessage` does not contain files. As we need to parse files for one of our features, we need to add a list of `SlackFile`s to this class.


slack-base/src/main/java/com/hubspot/slack/client/models/events/SlackEventMessageExtendedBase.java
```diff
@@ -0,0 +1,23 @@
+package com.hubspot.slack.client.models.events;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.hubspot.slack.client.methods.interceptor.HasUser;
+import java.util.Optional;
+
+public abstract class SlackEventMessageExtendedBase
```
zmarushchak-hs, line 7: @ykaliukh what does `Extended` mean in the model name?

ykaliukh: @zmarushchak-hs I changed the entire logic, please, check once more.

zmarushchak-hs (approved):

PR #338, opened 2024-04-26
Add support for deserializing Slack Connect files

djustinf:
The current SlackFileDeserializer doesn't handle Slack Connect files, which are described [here in the docs](https://api.slack.com/events/message/file_share#slack_connect_files). This change adds support for these files.


jaredstehler (approved):
🚢

PR #359, opened 2024-10-11
Add file info method to retrieve all file info

opaliukh:
Add file info method to retrieve all file info


ykaliukh (approved):
LGTM!
Maybe it is better to rename models because we only get 1 file. FilesInfoParamsIF -> FileInfoParamsIF

PR #380, opened 2025-05-07
Add markdown block type

lmorduch:
This adds support for the markdown block type (https://api.slack.com/reference/block-kit/blocks#markdown) - Want this to take advantage of more generic markdown syntax and a larger api limit for characters (12k)


lmorduch:
This is essentially a small piece of https://github.com/HubSpot/slack-client/pull/379/files broken out

PR #390, opened 2025-09-18
Add multi users select input

ethanez713:
Add support for the [MultiUsersSelect](https://docs.slack.dev/tools/node-slack-sdk/reference/types/interfaces/MultiUsersSelect/) input type. 

Tested locally and slack-client was able to parse values from this input type.


ethanez713:
Summarizing our slack convo:
I had to find the `selected_users` field from experimentation, I couldn't find this in docs either. I confirmed from our QA slack instance that that field is in fact present upon form submission!

PR #393, opened 2025-09-29
Add models to deal with Slack Work Object Flexpane and Action Buttons for it

Viktor-Pelepiak:
Description:
We, as the HubSpot Slack integration team, are continuing work on upgrading the link unfurling logic. We already support Work Objects and are going to extend this functionality further.
In this PR, I’ve added several models that will be used to add action buttons for Slack Work Objects and Flexpane, as well as to build the Work Object Flexpane itself. This PR also includes updates to the SlackClient to enable sending requests to the new entity.presentDetails Slack method.

Related issue: https://git.hubteam.com/HubSpot/slack-integration/issues/6303

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Set up JDK
       uses: actions/setup-java@v4
       with:
-        distribution: 'temurin'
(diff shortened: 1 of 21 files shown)


yyakhnytsia (approved):
LGTM.

PR #401, opened 2025-10-15
Fix/update work object models

Viktor-Pelepiak:
In this PR i've updated some models to fix serialization/deserialization bugs, and make models more safe for creation:
- Added enum for the `tag_color` fields prperty
- Added missing deserializers
- Added annotations to omit empty fields during serialization


yyakhnytsia (approved):
LGTM.

PR #440, opened 2026-04-15
Add new blocks to models

yyakhnytsia:
Add new block types: TaskCardBlock, MarkdownBlock and it's underlying structure to the models.


slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/richtextelements/RichTextObject.java
```diff
@@ -0,0 +1,16 @@
+package com.hubspot.slack.client.models.blocks.elements.richtextelements;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.hubspot.slack.client.models.blocks.elements.BlockElement;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes(
```
opaliukh, line 9: Should we also include rich_text_preformatted and rich_text_quote? What will happen if we receive a type that we do not support (if that’s somehow possible)? I am cautious about introducing logic-breaking changes.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/richtextelements/RichTextListIF.java
```diff
@@ -0,0 +1,28 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.google.common.collect.ImmutableList;
+import com.hubspot.immutables.style.HubSpotStyle;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Immutable;
+
+@Immutable
+@HubSpotStyle
+@JsonNaming(SnakeCaseStrategy.class)
+public interface RichTextListIF extends RichTextObject {
```
opaliukh, line 15: Can we add `indent`, `offset`, and `border`? It looks like these are missing, but adding them could be helpful in the future and would make our implementation consistent with the Slack documentation.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/MarkdownBlock.java
```diff
@@ -0,0 +1,84 @@
...
+    } else {
+      this.text = text;
+    }
+  }
+
+  public static MarkdownBlock of(String text) {
+    return new MarkdownBlock(text);
+  }
+
+  public static ImmutableList<Block> ofLongText(String text) {
```
opaliukh: > The Slack docs say the 12,000 character limit is cumulative across all markdown blocks in a single payload — not per block.
So if you call ofLongText with a 30K string, it could produce 3 blocks of ~10K each = 30K total in one payload. Slack would reject that because the cumulative limit across all markdown blocks is 12K.

Could you please verify this behavior?

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/UrlSourceIF.java
```diff
@@ -0,0 +1,30 @@
+package com.hubspot.slack.client.models.blocks.elements;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.hubspot.immutables.style.HubSpotStyle;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Immutable;
+
+/**
+ * Slack's Rich text block displays formatted, structured representation of text..
```
opaliukh: Please update it with the UrlSourceIF object information.

yyakhnytsia:
@opaliukh Fixed all comments.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/MarkdownBlock.java
yyakhnytsia: Yep, it fails on 13k+ characters in all blocks.

yyakhnytsia: I decided to split the interface from the creation logic, so we can handle limits separately in the service, which builds the palyload before sending.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/richtextelements/RichTextObject.java
yyakhnytsia: @opaliukh added

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/elements/richtextelements/RichTextListIF.java
```diff
@@ -0,0 +1,35 @@
...
+    return TYPE;
+  }
+
+  @JsonProperty("elements")
+  ImmutableList<RichTextSection> getElements();
+
+  @Value.Parameter
+  RichTextListStyle getStyle();
+
+  Optional<Integer> getIntend();
```
opaliukh:
```suggestion
  Optional<Integer> getIndent();
```

yyakhnytsia: @opaliukh fixed.

opaliukh (approved):

PR #451, opened 2026-05-27
Add new type of block ContextActionsBlock, fix issue with BlocksChunk.

yyakhnytsia:
Add new type of blocks: ContextActionsBlock from the Slack doc: https://docs.slack.dev/reference/block-kit/blocks/context-actions-block/. With the related nested components. Fix issue with the appendStream payload.


Viktor-Pelepiak (approved):
LGTM

PR #456, opened 2026-06-17
Add card block support

VictorHLi404:
## Summary

- Adds `CardIF` for Slack's [card block](https://docs.slack.dev/reference/block-kit/blocks/card-block), following the same pattern as `vili/table-block-support`
- Adds `SlackIconObjectIF` as a new composition object for the card's `slack_icon` field
- Reuses existing `elements.Image` for `hero_image`/`icon` and `Text` for text fields
- `actions` is modeled as `Optional<List<BlockElement>>` (flat array, not a wrapped Actions block)
- Registers `Card` in `Block.java`'s `@JsonSubTypes` and adds card-specific limits to `BlockElementLengthLimits`

## Test plan

- [ ] `CardBlockTest` covers round-trip deserialization for all 5 payload shapes (slack_icon+actions, hero_image, icon+actions, title-only, actions-only)
- [ ] Validation tests assert failures for: missing required fields, `icon`+`slack_icon` together, title >150 chars, body >200 chars, actions >3 elements

🤖 Generated with [Claude Code](https://claude.com/claude-code)


slack-base/src/main/java/com/hubspot/slack/client/models/blocks/CardIF.java
```diff
@@ -0,0 +1,112 @@
...
+public interface CardIF extends Block {
+  String TYPE = "card";
+
+  @Override
+  @Value.Derived
+  default String getType() {
+    return TYPE;
+  }
+
+  Optional<Image> getHeroImage();
```
yyakhnytsia, line 31: If you use optional fields in the class related to the Slack API payload, make sure it has `@JsonInclude(JsonInclude.Include.NON_ABSENT)`. Otherwise, when using this interface without some optional fields, you'll get invalid_arguments from Slack, without any details on what is wrong.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/CardIF.java
```diff
@@ -0,0 +1,112 @@
...
+
+  Optional<Text> getTitle();
+
+  Optional<Text> getSubtitle();
+
+  Optional<Text> getBody();
+
+  Optional<Text> getSubtext();
+
+  Optional<List<BlockElement>> getActions();
```
yyakhnytsia: Do we need an optional List here? As I know, we can just leave an empty collection

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/objects/SlackIconObjectIF.java
```diff
@@ -0,0 +1,22 @@
...
+public interface SlackIconObjectIF extends CompositionObject {
+  String TYPE = "icon";
+
+  @Value.Derived
+  default String getType() {
+    return TYPE;
+  }
+
+  @Value.Parameter
+  String getName();
```
yyakhnytsia: Slack icon has a list of allowed values:
https://docs.slack.dev/reference/block-kit/composition-objects/slack-icon-object
Could you please specify an enum instead, or at least leave a link to the documentation

yyakhnytsia (commented):
Reviewed, left comments.

PR #460, opened 2026-07-08
Make DataTable caption optional 

tonyhu1503:
Changed `DataTableIF#getCaption()` return type from `String` to `Optional<String>` 

Reason:
One of our team's worker has dependency on slack client and it started to report erros with DataTableIF saying
```
Caused by: com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `com.hubspot.slack.client.models.blocks.DataTable`, problem: Cannot build DataTable, some of required attributes are not set [caption]
```

[link to logFetch Errors](https://private.hubteam.com/errors/view/deployable/DevExMetricsTQ2-slack-events-message-worker/digest/8fea4d826ceabde74369d87b5ad8b7dd?timeRange=1782691200000/1783295940000)

related to https://github.com/HubSpotEngineering/DevExMetrics/issues/3216

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/DataTableIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/DataTableIF.java
@@ -28,7 +28,7 @@ public interface DataTableIF extends Block {
   }
 
   @Value.Parameter(order = 1)
-  String getCaption();
+  Optional<String> getCaption();
 
   @Value.Parameter(order = 2)
   ImmutableList<ImmutableList<DataTableCell>> getRows();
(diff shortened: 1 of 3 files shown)


tonyhu1503:
@opaliukh I heard from @VictorHLi404 that your team manages this repo, if you're ok with this change could i merge this PR please? thanks in advance

opaliukh:
@tonyhu1503 
Thanks for the fix! The deserialization issue makes sense — Slack sends `data_table` blocks without caption despite their docs listing it as required.

Just a heads up: changing the return type from String to Optional<String> is a breaking API change. I checked and no one is currently using DataTable outside of slack-client, so it’s safe for now. But worth keeping in mind for the future — if this model gets wider adoption, this kind of change would break downstream consumers.

Looks good to me otherwise, the test coverage is solid.

opaliukh (approved):

opaliukh (changes requested):
One more item: according to the Slack documentation, a caption is required.

https://docs.slack.dev/reference/block-kit/blocks/data-table-block

Image

opaliukh:
Looks like this is not used, so the difference with the Slack documentation may not be a real issue, as Slack doesn't provide this data.

tonyhu1503:
I see, thank you! it's weird that it's a required field but slack didn't provide it and caused the error. do you think this is a correct fix then?

opaliukh:
Looks correct to me. It looks like it’s not used by any downstream consumers outside of slack-client, so there’s no risk of breaking anyone.

opaliukh (approved):

PR #462, opened 2026-07-17
Add container block support

DeirdreCleary:
## Why

The Slack Block Kit API introduced an official `container` block type. This adds support for deserializing and building container blocks in this library, following the same patterns as other recently-added block types (card, task_card, data_table, table).

## What changed

- `ContainerBlockIF` / `ContainerBlock` — Immutables implementation of the `container` block with all Slack API fields: required `title` (Text) and `child_blocks` (ImmutableList<Block>); optional `subtitle`, `icon` (Image element), `width`, `is_collapsible`, `default_collapsed`, and `block_id`
- `ContainerBlockWidth` — enum for the `width` field values (`narrow`, `standard`, `wide`, `full`)
- `BlockElementLengthLimits` — added `MAX_CONTAINER_CHILD_BLOCKS(10)`
- `Block.java` — registered `ContainerBlock` in `@JsonSubTypes`
- Uses `@JsonInclude(NON_ABSENT)` (not `NON_EMPTY`) on optional fields, per pattern established in PR #456 to avoid Slack returning `invalid_arguments`

## Validation

`@Check` enforces:
- `title` text ≤ 150 characters
- `subtitle` text ≤ 150 characters (when present)
- `child_blocks` count ≤ 10

## Test plan

- `ContainerBlockTest` covers deserialization of all fields, minimal block, round-trip serialization, and all three validation failure cases
- `mvn test` passes locally for the `slack-base` module

## BRAVE

- **Backwards compatible** — adding a new `@JsonSubTypes.Type` entry is additive; existing code deserializing blocks is unaffected
- **Rollout/Rollback** — library change, no deploy; consumers upgrade at their own pace
- **Automated testing** — 7 unit tests added
- **Verification** — build passed locally
- **Expect failures** — none expected in existing consumers; new type is purely additive

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContainerBlockIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContainerBlockIF.java
@@ -8,7 +8,9 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.hubspot.immutables.style.HubSpotStyle;
 import com.hubspot.slack.client.models.blocks.elements.Image;
+import com.hubspot.slack.client.models.blocks.elements.richtextelements.RichTextBlock;
 import com.hubspot.slack.client.models.blocks.objects.Text;
+import com.hubspot.slack.client.models.blocks.objects.TextType;
 import java.util.Optional;
 import org.immutables.value.Value;
 import org.immutables.value.Value.Check;
@@ -17,7 +19,7 @@ import org.immutables.value.Value.Immutable;
 @Immutable
 @HubSpotStyle
 @JsonNaming(SnakeCaseStrategy.class)
-@JsonInclude(JsonInclude.Include.NON_EMPTY)
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
 public interface ContainerBlockIF extends Block {
   String TYPE = "container";
 
@@ -27,7 +29,9 @@ public interface ContainerBlockIF extends Block {
     return TYPE;
   }
 
-  Text getTitle();
+  Optional<Text> getTitle();
+
+  Optional<RichTextBlock> getRichTextTitle();
 
   ImmutableList<Block> getChildBlocks();
 
@@ -36,6 +40,8 @@ public interface ContainerBlockIF extends Block {
   Optional<Image> getIcon();
 
   Optional<ContainerBlockWidth> getWidth();
+
+  Optional<Boolean> getHasHeaderDivider();
 
   @JsonProperty("is_collapsible")
   Optional<Boolean> isCollapsible();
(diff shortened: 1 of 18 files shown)


slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContainerBlockIF.java
```diff
@@ -0,0 +1,70 @@
...
+public interface ContainerBlockIF extends Block {
+  String TYPE = "container";
+
+  @Override
+  @Value.Derived
+  default String getType() {
+    return TYPE;
+  }
+
+  Text getTitle();
```
yyakhnytsia: According to the [documentation](https://docs.slack.dev/reference/block-kit/blocks/container-block/) title is optional:
> Optional (one of title or rich_text_title is required) 

Also, it would be nice to have the other optional fields from the Slack documentation to have the full functionality of this component.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContainerBlockIF.java
```diff
@@ -0,0 +1,70 @@
...
+
+  ImmutableList<Block> getChildBlocks();
```
yyakhnytsia: Could you please check the supported type blocks: https://docs.slack.dev/reference/block-kit/blocks/container-block/#child-blocks
The used block has more types than are supported by the container type.
Also if we use the Block component it is possible to use recursive blocks: ContainerBlock inside ContainerBlock as a child component; not sure if such a structure is allowed.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContainerBlockIF.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @JsonProperty("is_collapsible")
+  Optional<Boolean> isCollapsible();
+
+  @JsonProperty("default_collapsed")
+  Optional<Boolean> isDefaultCollapsed();
+
+  @Check
+  default void check() {
+    Preconditions.checkState(
```
yyakhnytsia, line 54: Text type validation is missing:

> Title of the container, using plain_text formatting

We have two types of Text: plain_text and Markdown

yyakhnytsia (commented):
reviewed, left few comments.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContainerBlockIF.java
DeirdreCleary: Good point! `title` is now `Optional<Text>`, and I've added `rich_text_title` (`Optional<RichTextBlock>`) with a `@Check` validation that at least one of the two is present.

I also went through the full spec and added the remaining optional fields: `subtitle`, `icon`, `width`, `is_collapsible`, `default_collapse` and `has_header_divider`.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContainerBlockIF.java
DeirdreCleary: Fixed. We now validate in `@Check` that when `title` is present, its `TextType` is `PLAIN_TEXT` — building with a `mrkdwn` title throws `IllegalStateException: title must use plain_text formatting`.

slack-base/src/main/java/com/hubspot/slack/client/models/blocks/ContainerBlockIF.java
DeirdreCleary: Thanks for flagging this. I've looked at a few approaches for constraining child_blocks:

Option A — dedicated union interface (`ContainerChildBlock`) following the `BlockElement`/`RichTextObject` pattern. This enforces the constraint at compile time, but requires every supported block type’s IF interface to extend `ContainerChildBlock`, and both `Block.java` and `ContainerChildBlock.java` would need updating whenever a new block type is added. My concern is that this would become a maintenance burden over time.

Option B — allowlist in `@Check` using a `Set<String>` of the allowed `TYPE` constants (e.g. `Actions.TYPE`, `Divider.TYPE`, etc.), defined in `ContainerBlockIF`. The constraint lives in one place, unknown types (future Slack block types that deserialize as `UnknownBlock`) pass through for forward compatibility, and disallowed known types (e.g. `card`) are caught at build time with a clear error.

I'm inclined to go with Option B, but wanted to check your thoughts.

yyakhnytsia: If you have concerns about option A, then option B is also the case. For me, Option A is better for development, since we have a clear restriction.

DeirdreCleary: Updated with A

yyakhnytsia (approved):
LGTM.

yyakhnytsia (approved):
LGTM.

PR #464, opened 2026-07-28
Make threadTs optional for SetSuggestedPromptsParams

opaliukh:
Slack made threadTs optional for SetSuggestedPromptsParams [Slack doc ](https://docs.slack.dev/reference/methods/assistant.threads.setSuggestedPrompts/) as in the Agent view, we don't have threadTs.


yyakhnytsia (approved):
LGTM.

PR #466, opened 2026-07-29
Add @JsonInclude(NON_ABSENT) to optional fields in SetSuggestedPromptsParams

opaliukh:
Without this annotation, Optional.empty() fields are serialized as null in the JSON payload, which the Slack API may reject or misinterpret.

### DIFF
--- a/slack-base/src/main/java/com/hubspot/slack/client/methods/params/assistant/SetSuggestedPromptsParamsIF.java
+++ b/slack-base/src/main/java/com/hubspot/slack/client/methods/params/assistant/SetSuggestedPromptsParamsIF.java
@@ -1,5 +1,6 @@
 package com.hubspot.slack.client.methods.params.assistant;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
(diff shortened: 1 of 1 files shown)


slack-base/src/main/java/com/hubspot/slack/client/methods/params/assistant/SetSuggestedPromptsParamsIF.java
```diff
@@ -17,8 +18,10 @@ default String getChannel() {
     return getChannelId();
   }
 
+  @JsonInclude(JsonInclude.Include.NON_ABSENT)
```
yyakhnytsia: What if we put this annotation in the class level?

yyakhnytsia (commented):
reviewed, left comment.

slack-base/src/main/java/com/hubspot/slack/client/methods/params/assistant/SetSuggestedPromptsParamsIF.java
opaliukh: updated

yyakhnytsia (approved):
LGTM.

